PR #21, opened 2016-12-09
Bug 1309197: Add taskcluster-worker support; p=wcosta/dustin

djmitche:
taskcluster-worker runs via a launchd agent. It runs as root because it
needs root access to dscl command.

### DIFF
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -6,6 +6,15 @@
 
 # linux64 and OS X
 
+node /t-yosemite-r7-00[45][0-9]\.test\.releng\.scl3\.mozilla\.com/ {
+    $pin_puppet_server = "releng-puppet2.srv.releng.scl3.mozilla.com"
+    $pin_puppet_env = "dmitchell"
+    $aspects = [ 'low-security' ]
+    $slave_trustlevel = 'try'
+    $taskcluster = true
+    include toplevel::worker::releng::test::gpu
+}
+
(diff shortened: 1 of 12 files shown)


manifests/moco-nodes.pp
```diff
@@ -6,6 +6,15 @@
 
 # linux64 and OS X
 
+node /t-yosemite-r7-00[45][0-9]\.test\.releng\.scl3\.mozilla\.com/ {
+    $pin_puppet_server = "releng-puppet2.srv.releng.scl3.mozilla.com"
+    $pin_puppet_env = "dmitchell"
+    $aspects = [ 'low-security' ]
+    $slave_trustlevel = 'try'
+    $taskcluster = true
```
walac: This is not needed anymore.

modules/taskcluster_worker/templates/taskcluster-worker.yml.erb
```diff
@@ -0,0 +1,48 @@
...
+  workerId:         <%= @hostname %>
+  engine:           macosx
+  engines:
+    macosx:
+      createUser: true
+      sudo: true
+      userGroups: ['staff', 'everyone', 'localaccounts', '_developer', '_lpoperator', 'com.apple.sharepoint.group.1']
+  logLevel:         info
+  plugins:
+    disabled:       ['success', 'interactive']
```
walac: `success` plugin should not be removed, my mistake.

PR #34, opened 2017-07-12
bug 1374963 test

nthomas-mozilla:
For a travis test only (linting etc)


nthomas-mozilla:
Travis clean, landing will be in hg as per usual

PR #35, opened 2017-09-12
Add Fennec puppet changes for release runner integration.

MihaiTabara:


MihaiTabara:
This is for ease of review and testing purposes (for automation tests).
Once this is green flagged and r+, it'll follow the mercurial path as usual.

modules/releaserunner2/manifests/init.pp
```diff
@@ -0,0 +1,98 @@
...
+            require   => Python::Virtualenv[$releaserunner2::settings::root],
+            mode      => '0600',
+            owner     => $users::builder::username,
+            group     => $users::builder::group,
+            content   => secret('funsize_signing_pvt_key'),
+            show_diff => false;
+    }
+
+    mercurial::repo {
+        'releaserunner-tools':
```
rail: Can you rename it to `releaserunner2-tools` or something. Otherwise it's not unique, so you cannot run releaserunner and releaserunner2 on the same node.

MihaiTabara: Woot, good catch!

MihaiTabara:
Pushed via mercurial, we can close this.

PR #45, opened 2018-06-05
Require gcc to be installed before creating buildslave virtualenv.

mozbhearsum:
I have no idea why this started to happen all of a sudden. Looking over old logs shows Twisted installing just fine without gcc. In any case, this should fix it.

### DIFF
--- a/modules/buildslave/manifests/install.pp
+++ b/modules/buildslave/manifests/install.pp
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 class buildslave::install {
-    if ($::operatingsystem != windows) {
+    if ($::operatingsystem == CentOS) {
         include packages::gcc
     }
     include packages::mozilla::python27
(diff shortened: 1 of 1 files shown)


modules/buildslave/manifests/install.pp
```diff
@@ -2,6 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 class buildslave::install {
+    if ($::operatingsystem != windows) {
```
lundjordan: judging by https://dxr.mozilla.org/build-central/source/puppet/modules/packages/manifests/gcc.pp#14

should we invert this and make it `if ($::operatingsystem == CentOS)`?

modules/buildslave/manifests/install.pp
```diff
@@ -18,6 +21,14 @@
 
     $py_require = Class['packages::mozilla::python27']
 
+    $virtualenv_require = $::operatingsystem ? {
+        windows => [Class['packages::mozilla::python27']],
```
lundjordan: similar to comment above, should we define CentOS as the exception and windows, ubuntu, etc the `default`?

mozbhearsum:
It's late here - so even if this is approved I don't intend to merge it until tomorrow. If someone else would like to, feel free.

modules/buildslave/manifests/install.pp
```diff
@@ -18,6 +21,14 @@
 
     $py_require = Class['packages::mozilla::python27']
```
lundjordan: is this still needed?

PR #49, opened 2018-06-05
Add thunderbird shipit scriptworkers.

tomprince:
Depends on https://github.com/mozilla-releng/shipitscript/pull/7

### DIFF
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -1142,6 +1142,22 @@ node /^shipitworker-.*\.srv\.releng\..*\.mozilla\.com$/ {
     include toplevel::server::shipitscriptworker
 }
 
+node /^tb-shipit-dev-\d+\.srv\.releng\..*\.mozilla\.com$/ {
+    $aspects                  = [ 'maximum-security' ]
+    $shipit_scriptworker_env  = 'tb-dev'
+    $timezone                 = 'UTC'
+    $only_user_ssh            = true
+    include toplevel::server::shipitscriptworker
+}
+
+node /^tb-shipit-\d+\.srv\.releng\..*\.mozilla\.com$/ {
+    $aspects                  = [ 'maximum-security' ]
+    $shipit_scriptworker_env  = 'tb-prod'
+    $timezone                 = 'UTC'
+    $only_user_ssh            = true
+    include toplevel::server::shipitscriptworker
+}
+
 # Treescript workers
 node /^treescriptworker-dev\d*\.srv\.releng\..*\.mozilla\.com$/ {
     $aspects          = [ 'maximum-security' ]
--- a/modules/shipit_scriptworker/manifests/settings.pp
+++ b/modules/shipit_scriptworker/manifests/settings.pp
@@ -29,6 +29,7 @@ class shipit_scriptworker::settings {
         sign_chain_of_trust      => false,
         verify_chain_of_trust    => true,
         verify_cot_signature     => false,
+        cot_product              => 'firefox',
 
         ship_it_instances        => {
             "${ship_it_stage_instance_scope}" => $ship_it_stage_instance_config,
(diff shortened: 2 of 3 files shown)


manifests/moco-nodes.pp
```diff
@@ -1107,6 +1107,22 @@
     include toplevel::server::shipitscriptworker
 }
 
+node /^tb-shipit-dev-\d+\.srv\.releng\..*\.mozilla\.com$/ {
```
MihaiTabara: Nit, not blocking: I think it'd be good to be consisent with these namings. Reading https://hg.mozilla.org/build/puppet/file/default/manifests/moco-nodes.pp I see lots of inconsistencies - but not necessarily for TB, but in general. In this case, I see:

tb-signing
tb-depsigning-worker
tb-balrog-
^tb-balrogworker-dev\
^tb-beetmover-
^tb-beetmover-dev

So we'd rather have them be:
`tb-shipit` (prod)
`tb-shipitworker-dev` 

But I see your set follows beetmover so I guess afterall there's some consistency. 
My own preferences would be to keep the "worker" string within these FQDN to make it clear upfront they are all scriptworkers under the hood.

modules/shipit_scriptworker/files/requirements.txt
```diff
@@ -29,4 +29,4 @@ taskcluster==3.0.1
 urllib3==1.22
 virtualenv==16.0.0
 yarl==1.2.4
-shipitscript==1.0.0  # puppet: nodownload
+shipitscript==2.0.0  # puppet: nodownload
```
MihaiTabara: FYI: I might have rolled-out some 2.0.0-dev work for this when testing with https://github.com/mozilla-releng/shipitscript/pull/8/files the other day. Sorry for that. I won't rollout that PR for another 2-3 weeks so feel free to rollout this and I'll bump again when the time comes.

modules/shipit_scriptworker/manifests/settings.pp
```diff
@@ -18,17 +18,19 @@
         password                  => secret('shipit_scriptworker_ship_it_password_dev'),
     }
 
-    $_env_configs             = {
+    $env_config = {
       'dev'  => {
         worker_group             => 'shipit-dev',
         worker_type              => 'shipit-dev',
         verbose_logging          => true,
```
MihaiTabara: I think this is unused since in modules/shipit_scriptworker/manifests/init.pp you're using the one from line 111 - https://github.com/mozilla/build-puppet/pull/49/files#diff-16fc35f7191a6528caf61089d8af4f3eR111

modules/shipit_scriptworker/manifests/settings.pp
```diff
@@ -40,10 +42,12 @@
...
```
MihaiTabara: I think this is unused since in modules/shipit_scriptworker/manifests/init.pp you're using the one from line 111 - https://github.com/mozilla/build-puppet/pull/49/files#diff-16fc35f7191a6528caf61089d8af4f3eR111

modules/shipit_scriptworker/manifests/settings.pp
```diff
@@ -55,32 +59,56 @@
             },
         },
       },
+      'tb-dev'  => {
+        worker_group             => 'shipit-v1',
+        worker_type              => 'tb-shipit-dev',
+        verbose_logging          => true,
```
MihaiTabara: I think this is unused since in modules/shipit_scriptworker/manifests/init.pp you're using the one from line 111 - https://github.com/mozilla/build-puppet/pull/49/files#diff-16fc35f7191a6528caf61089d8af4f3eR111

modules/shipit_scriptworker/manifests/settings.pp
```diff
@@ -55,32 +59,56 @@
...
+        cot_product              => 'thunderbird',
+
+        ship_it_instances        => {
+            "project:comm:thunderbird:releng:ship-it:staging" => $ship_it_stage_instance_config,
+        },
+      },
+      'tb-prod'  => {
+        worker_group             => 'shipit-v1',
+        worker_type              => 'tb-shipit-prod',
+        verbose_logging          => true,
```
MihaiTabara: I think this is unused since in modules/shipit_scriptworker/manifests/init.pp you're using the one from line 111 - https://github.com/mozilla/build-puppet/pull/49/files#diff-16fc35f7191a6528caf61089d8af4f3eR111

manifests/moco-nodes.pp
tomprince: One thing to keep in mind (I ran into this with the treescriptworker), is that there is a limit of 22 characters for worker ids, which we take from the hostname. I do agree that we should be more consistent in general with these names.

I'd suggest, though, that if we want `worker` in the name, we should add another level of name and put it there.

modules/shipit_scriptworker/files/requirements.txt
tomprince: Shouldn't conflict. `-dev` will only be pulled in if asked for explicitly.

manifests/moco-nodes.pp
MihaiTabara: Ah, damn, I forgot of that, mea culpa. You're right, +1 to your suggestion.

modules/shipit_scriptworker/files/requirements.txt
MihaiTabara: Yeah, I agree. I mentioned it just in case you're wondering what a 2.0.0-devX package is doing there :)

lundjordan:
@tomprince should this have landed?

PR #61, opened 2018-06-07
Upgrade as much of releasetasks as we can

mozbhearsum:
I tested this on bm83 as much as possible (I had to hack out a bunch of checks to get as far as task submission). Notable problems/fixes:
* The latest taskcluster library doesn't support encrypted vars
* Had to install pyasn1 for pgpy
* Had to downgrade "six" for pgpy to work (see https://github.com/SecurityInnovation/PGPy/issues/217)

In the end, I was able to get a graph created. I can't guarantee 100% that this will work in production because of all the hacks that I did, but I think this is as well tested as it's going to get.

### DIFF
--- a/modules/releaserunner/files/requirements.txt
+++ b/modules/releaserunner/files/requirements.txt
@@ -1,37 +1,46 @@
 # python_version: 27
 # install six before cryptography
-six==1.9.0
-Fabric==1.5.1
-Jinja2==2.6
-PGPy==0.3.0
+six==1.11.0
+Fabric==2.1.3
+Jinja2==2.10
+PGPy==0.4.3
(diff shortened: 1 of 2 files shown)


mozbhearsum:
I couldn't find any evidence that installing build/tools was necessary. I tested this with a full rebuild of /builds/releaserunner, and it worked fine without it. If you're referring to all of the imports that release-runner.py does, note that it appends sys.path before doing so (https://github.com/mozilla/build-tools/blob/master/buildfarm/release/release-runner.py#L21). 

I could be missing something here though.

escapewindow:
Ah, then we're probably good.

PR #66, opened 2018-06-08
bug 1459005: Add puppet module/manifests for mergeday instance

mozbhearsum:
With a machine puppetized against this I was able to run the no-op merges without issue.


manifests/moco-nodes.pp
```diff
@@ -342,6 +342,13 @@
     include toplevel::server::buildduty_tools
 }
 
+# mergeday
+
+node /^mergeday\d+\.srv\.releng\.(use1|usw2)\.mozilla\.com$/ {
+    $aspects = [ 'medium-security' ]
```
MihaiTabara: Hm, I didn't dig into the specifics of what this encompasses, but is there any reason why this isn't `high-security`?

modules/mergeday/manifests/install.pp
```diff
@@ -0,0 +1,25 @@
...
+    include ::config
+    include users::buildduty
+    include packages::mercurial
+    include packages::mozilla::python27
+    include packages::mozilla::py27_virtualenv
+    include packages::mozilla::py27_mercurial
+    include packages::patch
+
+    python::virtualenv {
+        "/home/buildduty/mergeday":
```
MihaiTabara: Nit, not blocking: I thought we're retiring this terminology in favor of `ciduty` 😄 Or better said `releaseduty`?

modules/mergeday/manifests/install.pp
```diff
@@ -0,0 +1,25 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+class mergeday::install {
+    include ::config
+    include users::buildduty
+    include packages::mercurial
+    include packages::mozilla::python27
+    include packages::mozilla::py27_virtualenv
+    include packages::mozilla::py27_mercurial
+    include packages::patch
```
MihaiTabara: Hm, unclear to me, why do we need this?

manifests/moco-nodes.pp
```diff
@@ -342,6 +342,13 @@
     include toplevel::server::buildduty_tools
 }
 
+# mergeday
+
+node /^mergeday\d+\.srv\.releng\.(use1|usw2)\.mozilla\.com$/ {
```
MihaiTabara: As far as I can tell from AWS console, we have a `mergeday1` in the `us-west-2` and your [warrior PR](https://github.com/mozilla-releng/releasewarrior-2.0/pull/150/files#diff-f188716043a5f41d82e523895637b90aR5) confirms that. Is there any reason we're allowing here `use1` as well? Or are we keeping this for security reasons in case us-west-2 is down and we quickly need to rely on a newly ramped up `us-east-1` new instance?

manifests/moco-nodes.pp
mozbhearsum: I could change it, I guess. There's no secrets or other valuable information on this machine though - so I don't think it matters.

tomprince: A ssh-agent connection could probably be considered valuable?

mozbhearsum: That's true. Eh, let's mark this high security then.

modules/mergeday/manifests/install.pp
mozbhearsum: We are, but I don't want to block having a proper merge day instance on unrelated work like that.

modules/mergeday/manifests/install.pp
mozbhearsum: It's used as part of the migration scripts: https://hg.mozilla.org/mozilla-central/file/2cca333f546f38860f84940d4c72d7470a3410f4/testing/mozharness/scripts/merge_day/gecko_migration.py#l230

manifests/moco-nodes.pp
mozbhearsum: I wrote it like this to leave room for future expansion. I can hardcode a single hostname here if you prefer.

MihaiTabara: Nope, makes sense. Just wanted to confirm my understanding was accurrate. Let's leave it like that, it doesn't hurt.

modules/mergeday/manifests/install.pp
MihaiTabara: Ah, I totally missed that, sorry. Thanks for the pointer!

modules/mergeday/manifests/install.pp
MihaiTabara: No worries, we can ignore for now and address later on.

PR #67, opened 2018-06-11
Rewrite puppet manifest locations to the mozilla-releng org on Github.

mozbhearsum:
This rewrites the remaining traces of https://hg.mozilla.org/build/puppet, and the original github location to the new, final home in the RelEng org.

@davehouse - I'm flagging you for review because of the changes to `puppet-darwin-run-puppet.sh.erb`. You'll notice that I've basically disabled the manifest version check for now. I'll need to make changes to the Puppet masters before we can enable it again. How important is that check?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 See https://wiki.mozilla.org/ReleaseEngineering/PuppetAgain.
 
-[![Build Status](https://travis-ci.org/mozilla/build-puppet.png)](https://travis-ci.org/mozilla/build-puppet)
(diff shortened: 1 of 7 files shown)


davehouse:
@mozbhearsum the manifest version check is not critical. In fact, about half of the time the network is not available yet and the check is not able to be performed. So, it is a "nice to have" but I don't think it gives us much value.

I think we can remove the version check and depend on elapsed time. We can always manually trigger puppet to run on machines if we have a change that we need to force across all workers quickly (and not wait for them to pick it up).

mozbhearsum:
OK, great! I'll just remove that check entirely instead of the half baked disabling, in that case.

davehouse:
+1 thank you. that was quick

PR #73, opened 2018-06-13
Bug1457003 roller add ssh user for workers

davehouse:
@dividehex @dragoscrisan 

1. create roller user account on mac workers (will add onto linux workers in a separate patch after I've tested if additions are needed)
2. restrict that user to the reboot command
3. allow the user to run reboot as root without a password
4. do not allow a password login for the roller user

### DIFF
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -18,6 +18,7 @@ node /^t-yosemite-r7-\d+\.test\.releng\.(mdc1|mdc2|scl3)\.mozilla\.com$/ {
     $slave_trustlevel = 'try'
     include fw::profiles::osx_taskcluster_worker
     include toplevel::worker::releng::generic_worker::test::gpu
+    include users::roller  # roller needs ssh to reboot macs
 }
 
 # Linux on moonshot in mdc1 running taskcluster worker
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -18,7 +18,6 @@ node /^t-yosemite-r7-\d+\.test\.releng\.(mdc1|mdc2|scl3)\.mozilla\.com$/ {
     $slave_trustlevel = 'try'
     include fw::profiles::osx_taskcluster_worker
     include toplevel::worker::releng::generic_worker::test::gpu
-    include users::roller  # roller needs ssh to reboot macs
 }
 
 # Linux on moonshot in mdc1 running taskcluster worker
--- a/modules/users/manifests/roller/account.pp
+++ b/modules/users/manifests/roller/account.pp
@@ -22,38 +22,7 @@ class users::roller::account($username, $group, $grouplist, $home) {
             # relevant fixes.
             # NOTE: this user is *not* an Administrator.  All admin-level access is granted via sudoers.
             case $::macosx_productversion_major {
-                '10.7': {
-                    if (secret('roller_pw_saltedsha512') == '') {
-                        fail('No roller password saltedsha512 set')
-                    }
-                    darwinuser {
-                        $username:
-                            gid      => $group,
-                            shell    => '/bin/bash',
-                            home     => $home,
-                            password => secret('roller_pw_saltedsha512'),
-                            comment  => 'roller',
-                            notify   => Exec['kill-roller-keychain'];
-                    }
-                    $user_req = Darwinuser[$username]
-                }
-                '10.8': {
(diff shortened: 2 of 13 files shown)


manifests/moco-nodes.pp
```diff
@@ -18,6 +18,7 @@
     $slave_trustlevel = 'try'
     include fw::profiles::osx_taskcluster_worker
     include toplevel::worker::releng::generic_worker::test::gpu
+    include users::roller  # roller needs ssh to reboot macs
```
davehouse: Copied from mozreview:
> I think we should be doing ssh on the linux taskcluster host in addition to the osx hosts.  A ssh->reboot is a much more graceful and cleaner shutdown.  If that fails, roller should fall back to issuing IPMI shutdown.  This is also the model slaveapi uses.
I'm assuming the IPMI soft-reboot doesn't signal the os to reboot, if it does, disregard.  The goal is a graceful shutdown and then falling back to a hard reboot.
5/2/2018 [:dividehex]

modules/users/manifests/roller/account.pp
```diff
@@ -0,0 +1,98 @@
...
+    # create the user
+
+    case $::operatingsystem {
+        Darwin: {
+            # use our custom type and provider, based on http://projects.puppetlabs.com/issues/12833
+            # This should be replaced with 'user' once we are running a version of puppet containing the
+            # relevant fixes.
+            # NOTE: this user is *not* an Administrator.  All admin-level access is granted via sudoers.
+            case $::macosx_productversion_major {
+                '10.7': {
```
davehouse: Copied from mozreview:
> We don't support 10.8 or 10.9 at all and I don't think we will ever run taskcluster on 10.7, therefore we can rip those case blocks out and only case for 10.10.
5/2/2018 [:dividehex]

modules/users/manifests/roller/account.pp
```diff
@@ -0,0 +1,98 @@
...
+                }
+                '10.9','10.10': {
+                    if (secret('roller_pw_pbkdf2') == '' or secret('roller_pw_pbkdf2_salt') == '') {
+                        fail('No roller password pbkdf2 set')
+                    }
+                    user {
+                        $username:
+                            shell      => '/bin/bash',
+                            home       => $home,
+                            password   => secret('roller_pw_pbkdf2'),
```
davehouse: Copied from mozreview:
> The roller users should *NOT* have a password at all since roller should only be accessing it via ssh key. It will just become another password that will have to be rotated at some point.
5/2/2018 [:dividehex]

modules/users/manifests/roller/setup.pp
```diff
@@ -0,0 +1,47 @@
...
+            sudoers::custom {
+                'roller-reboot':
+                    user    => $users::roller::username,
+                    runas   => 'root',
+                    command => "$sudoers::settings::rebootpath, /usr/sbin/bless";
+            }
+        }
+    }
+
+    $mac_netboot_ips = [ '10.26.52.17',  # install.build.releng.scl3
```
davehouse: Copied from mozreview:
> The netboot ip array probably belongs in ::config or hiera. It makes is easier to find when looking to change it, otherwise it becomes the type of thing we miss when setting up or decomming deploystudio servers.
5/2/2018 [:dividehex]

manifests/moco-nodes.pp
davehouse: Copied from mozreview:
> This is probably note the best place to include the user module. I would suggest putting it toplevel::worker and limit its application to at least Darwin, maybe Linux
5/2/2018 [:dividehex]

modules/users/manifests/roller/account.pp
davehouse: Copied from mozreview:
> I'm researching how to set up users without passwords in osx. I'll update with a new patch once I have that tested and working.
[:dhouse]

manifests/moco-nodes.pp
davehouse: Copied from mozreview:
> The ipmi soft-reboot issues an ACPI shutdown and so I think it is gracefully caught by the OS. But let's add the roller user to the linux boxes also; we'll need ssh to the linux machines for creating loaners anyway.
[:dhouse]

modules/users/manifests/roller/account.pp
davehouse: I've added a MatchUser entry to the sshd_config to prevent password login for the roller user. I did not find another good method to remove or avoid passwords on the roller user (removing it from the user appears to break things in osx).

PR #77, opened 2018-06-20
Bug 1460446 - Segregate Macs used for staging

dragoscrisan:
Move macs workers used for staging into another pool to not grab jobs from production.
What I do:
- introduced a node variable, called environment. This variable has the following values for the moment: production and staging. The default value is production.


manifests/moco-nodes.pp
```diff
@@ -1199,3 +1199,13 @@
     $aspects = [ 'low-security' ]
     include toplevel::server
 }
+
+node 't-yosemite-r7-380.test.releng.mdc1.mozilla.com' {
+    $aspects           = [ 'low-security' ]
+    $slave_trustlevel  = 'try'
+    #$pin_puppet_server = 'releng-puppet2.srv.releng.scl3.mozilla.com'
+    #$pin_puppet_env    = 'dcrisan'
```
petemoore: Let's delete these two commented out lines.

dragoscrisan: Removed the commented lines

PR #78, opened 2018-06-21
Update readme to verify Taskcluster checks work and pass

Callek:


Callek:
the python:XX-slim don't have git on them, so this turns out it is needed to land to fix that

PR #79, opened 2018-06-21
Bug1457003 roller linux ssh

davehouse:
@dividehex @dragoscrisan 

1. fix network entries for fw module to add roller to iptables without trying to resolve the ips (I think the network firewall is only allowing ingress)
2. add "roller" user to Ubuntu workers with sudo for reboot
3. accept command "reboot" and not the full path
4. decrease the timeout for a reboot's is_down and is_up; If a reboot takes longer than a minute for the shutdown, then there is a problem and we can try a forced shutdown (like with ipmi/snmp).


davehouse:
@dividehex r? some fixes I found (more testing before I merged)

PR #80, opened 2018-06-21
Update all scriptworker requirements files

escapewindow:
We both want to reduce the size of the pyup patch and update python-gnupg sooner.

I'm going to flag a lot of people for review; please comment on the scriptworker instances you know about.


escapewindow:
Looks like the new taskcluster check needs to use an image with git installed (or the task needs to install git)

Callek:
stamp for addonscript and treescript (please only merge after monday/tuesday though)

escapewindow:
Ok. If we need to merge earlier I can leave out pushapk, tree, and addonscript. Otherwise I'll wait til next week.

@JohanLorenzo can you let me know when we're clear to update pushapk? Going forward we'll likely accept the pyup PRs across the entire repo, so if we need to keep pushapk or other instances back we may need to pin them to user envs or mark them for pyup to ignore.

JohanLorenzo:
> Could you please hold to land the pushapk changes before https://bugzilla.mozilla.org/show_bug.cgi?id=1464034 is fixed?

The problem is fixed. I just updated pushapkworker to master and reran the last job. It failed because of a missing dep. I'll make a follow up patch. cc @escapewindow  

https://tools.taskcluster.net/groups/R3l-KgP1SuSC5XhAniGNxg/tasks/aFhE8qhQQ3S17Q0mS-1n5A/runs/1/logs/public%2Flogs%2Flive_backing.log#L27

PR #84, opened 2018-06-25
Update nrpe package on roller servers

dragoscrisan:
Install nrpe 3.0 to roller servers

### DIFF
--- a/modules/packages/manifests/setup.pp
+++ b/modules/packages/manifests/setup.pp
@@ -225,7 +225,7 @@ class packages::setup {
             }
             # to flush the package index, increase this value by one (or
             # anything, really, just change it).
-            $repoflag = 42
+            $repoflag = 43
             file {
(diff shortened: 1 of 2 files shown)


modules/packages/manifests/setup.pp
```diff
@@ -375,6 +375,11 @@
                     options      => ['arch=amd64'],
                     gpg_id       => '0EBFCD88',
                     gpg_key      => 'puppet:///modules/packages/0EBFCD88.txt';
+                'nrpe':
+                    url_path     => 'repos/apt/custom/nrpe',
+                    distribution => $::lsbdistcodename,
+                    options      => ['trusted=yes'],
```
davehouse: can we remove the "trusted=yes" option?

dragoscrisan: all repos are set with trused=yes options

PR #86, opened 2018-06-25
Change distinguished puppet master

dividehex:
This is a bit transitional since it will list both the new and old DM under the fw module.  Once the dust has settled, we can remove the old DM from the fw network DM list.


davehouse:
+1 nice alignment

PR #92, opened 2018-06-29
Bug 1471541 - update tooltool to 1c6943c

garbas:
reason for this is that new `tooltool.py` uses new url


garbas:
@markcor I hope it is ok to ping you for a review. I noticed you were reviewer last time when dustin was making this change, but that was 3 years ago.

garbas:
@rail would you be able to review this PR?

PR #100, opened 2018-07-03
 Bug 1461913 - Quarantine workers whose generic-worker process exit with code 69

dragoscrisan:
- check the status code returned by generic worker
- if status code is 69, quarantine worker and create the bug
- if the worker will be back to normal, then remove the worker from the quarantine and close the bug


modules/generic_worker/manifests/init.pp
```diff
@@ -11,6 +11,8 @@
             $macos_version = regsubst($::macosx_productversion_major, '\.', '')
             $taskcluster_client_id = secret('generic_worker_macosx_client_id')
             $taskcluster_access_token = hiera('generic_worker_macosx_access_token')
+            $quarantine_client_id = secret('quarantine_client_id')
+            $quarantine_access_token = hiera('qauarntine_access_token')
```
davehouse: typo "quarantine"

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
...
+export TASKCLUSTER_CLIENT_ID="<%= @quarantine_client_id %>"
+export TASKCLUSTER_ACCESS_TOKEN="<%= @quarantine_access_token %>"
+
+if [ $exitstatus -eq 69 ]; then
+    if [ -f /Users/cltbld/quarantined ]; then
+        echo "Worker is quarantined"
+    else
+        echo "Quarantine worker"
+        # quarantine the worker
+        /usr/local/bin/quarantine-worker releng-hardware "<%= @worker_type %>" "<%= @worker_group %>" "<%= @hostname %>" 720h5s
```
davehouse: what is the 72 argument, can that be a variable from puppet also (to expose it)

petemoore: > what is the 72 argument

I think 720 hours and 5 seconds (3 days and 5 seconds) - which is how long the worker gets quarantined for. This seems a little low - perhaps it should be an indefinite quarantine (e.g. 100 years), so that somebody has to look into it in order to remove it from quarantine.

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
 # wrapper script...
 /usr/local/bin/generic-worker "$@" 2>&1 | logger -t generic-worker -s
 
+exitstatus=( ${PIPESTATUS[0]} )
+
+echo $exitstatus
```
petemoore: I don't think we need this `echo`, the worker already logs its exit status. If you choose to keep it, please log that it is the generic-worker exit code (e.g. `echo "Generic worker exit code was ${exitstatus}"`).

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
 # wrapper script...
 /usr/local/bin/generic-worker "$@" 2>&1 | logger -t generic-worker -s
 
+exitstatus=( ${PIPESTATUS[0]} )
```
petemoore: I wasn't familiar with `PIPESTATUS` - nice job!

Nit: I suspect it doesn't require the outer brackets, e.g. maybe this is sufficient:
```
exitstatus=${PIPESTATUS[0]}
```

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
...
+
+export TASKCLUSTER_CLIENT_ID="<%= @quarantine_client_id %>"
+export TASKCLUSTER_ACCESS_TOKEN="<%= @quarantine_access_token %>"
+
+if [ $exitstatus -eq 69 ]; then
+    if [ -f /Users/cltbld/quarantined ]; then
+        echo "Worker is quarantined"
+    else
+        echo "Quarantine worker"
+        # quarantine the worker
```
petemoore: I prefer: `# quarantine the worker using github.com/mozilla-platform-ops/quarantine-worker CLI`.

The quarantine-worker project also needs a `README.md` to explain command line args and explain what it does, and why it is useful etc.

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
...
+        /usr/local/bin/quarantine-worker releng-hardware "<%= @worker_type %>" "<%= @worker_group %>" "<%= @hostname %>" 720h5s
+        # create the bug
```
petemoore: This will be awesome when implemented!

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
...
+
+        # create a semaphore file
+        echo "quarantined worker" > /Users/cltbld/quarantined
+        # create the bug
+    fi
+elif [ -f /Users/cltbld/quarantined ]; then
+
+    # if the worker was in quarantine, but now the exitcode is other that 69
+    # remove worker from quarantine
+    /usr/local/bin/quarantine-worker releng-hardware "<%= @worker_type %>" "<%= @worker_group %>" "<%= @hostname %>" 0s
```
petemoore: If the worker is in quarantine, it won't be given tasks from the queue, so it will likely exit with exit code 68 (idle timeout) which would not be a reason to take the worker out of quarantine.

I don't think any automated *check* should take the worker out of quarantine. The only way a worker should be taken out of quarantine automatically is if there is e.g. an automatic worker reimage process and there is some way to be sure there is not a hardware failure etc.

We probably also need an override feature so that it is possible to disable the auto-quarantine process on a given worker or set of workers, in case we have a rare intermittent issue that causes a 69 exit code but that is not really a worker failure.

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -14,6 +14,39 @@ rm -f tasks-resolved-count.txt
...
+
+    # close the bug
+
+    # remove the semaphore file
+
+    /bin/rm -rf /Users/cltbld/quarantined
```
petemoore: This whole `elif` section can go, I believe, for the reason above.

petemoore:
@dragoscrisan I don't have permission to rename the PR title, but can you prefix it with the bug number, e.g. __Bug 1461913 - Quarantine workers whose generic-worker process exit with code 69__ ?
Thanks!

modules/generic_worker/templates/run-generic-worker.sh.erb
dragoscrisan: petemoore: You are right. I used this echo for debug, I'll remove it

dragoscrisan: 720 hours and 5 seconds means 30 days and 5 second. I think is enough time to someone have a look and fix the worker

modules/generic_worker/templates/run-generic-worker.sh.erb
dragoscrisan: exitstatus=${PIPESTATUS[0]} sisn't work. you need to put ${PIPESTATUS[0]} between ()

modules/generic_worker/templates/run-generic-worker.sh.erb
dragoscrisan: +1

modules/generic_worker/templates/run-generic-worker.sh.erb
dragoscrisan: +1. This will ease my work. So I suggest to keep auto quarantine and auto generate the bug if the exit code 69 is present, but remove the worker from the quarantine and close the bug will be part of the CI duty task

modules/generic_worker/templates/run-generic-worker.sh.erb
```diff
@@ -10,10 +10,36 @@
...
 /usr/local/bin/generic-worker "$@" 2>&1 | logger -t generic-worker -s
 
+exitstatus=( ${PIPESTATUS[0]} )
+
+export TASKCLUSTER_CLIENT_ID="<%= @quarantine_client_id %>"
+export TASKCLUSTER_ACCESS_TOKEN="<%= @quarantine_access_token %>"
+
+if [ $exitstatus -eq 69 ]; then
+    if [ -f /Users/cltbld/quarantined ]; then
+        echo "Worker is quarantined"
```
petemoore: If `/Users/cltbld/quarantined` exists, it means on the previous run, the worker was quarantined. When this happens at this point in the code, we know the worker believes itself to be in an invalid bad/corrupted state after this new run. Therefore, this could be because either:

1) The failure mode is hit without claiming a task, e.g. on generic-worker initialisation. In this case, the machine will start boot-looping - that could be potentially harmful to the hardware if it is e.g. rebooting every minute or so. If several machines (e.g. potentially an entire worker pool) start doing this at the same time due to a code bug, I'm concerned this could be even more damaging, even to other machines in the data centre.

2) The worker quarantine was lifted or expired naturally, the queue provided a task for the worker, and during task execution the failure mode was discovered. In this case, we are not in a boot-looping situation, since the worker will be quarantined again, and will hit an idle timeout on the next run if the quarantine is not lifted, which occurs after 96 hours.

The long and short of this is that I think it would be wise to set a minimum worker uptime duration between worker reboots in the case when exit code 69 is detected. If a worker fails immediately with an error code 69, we shouldn't immediately reboot. There is a further argument to say once we hit exit code 69, the worker should just avoid the reboot entirely, and leave the machine running. The only reason to keep rerunning generic-worker would be to catch the case where the root cause is of the problem gets resolved and the quarantine is actively lifted, and we want the worker to immediately start taking tasks again. In this case, it removes the need for someone to reboot all affected workers.

A happy middle ground may be to limit rebooting to once per hour in the case of an exit code 69. Note, this restriction shouldn't apply for other exit codes, since e.g. if a worker runs a task that only lasts one minute, it should still reboot immediately.

modules/packages/manifests/mozilla/generic_worker.pp
```diff
@@ -33,6 +34,14 @@
                     owner  => root,
                     group  => wheel,
             }
+            # install quarantine-worker, Bug 1461913
+            file {
+                '/usr/local/bin/quarantine-worker':
```
petemoore: For generic-worker there is a comment about where binaries should be downloaded from. Can you add a similar comment for quarantine-worker?

lundjordan:
hi @dragoscrisan - there are a number of PRs (like this one) that are approved in build-puppet. Should any of them have merged. Gentle ping, feel free to ignore if all is well.

modules/generic_worker/manifests/control_bug.pp
```diff
@@ -0,0 +1,37 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This class is used to create and close bugs automatically
```
petemoore: _This class provides a utility to the worker, to enable it to raise a Bugzilla bug against itself, if it considers that it is in a bad state._

modules/generic_worker/manifests/init.pp
```diff
@@ -4,13 +4,16 @@
 
 class generic_worker {
     include packages::mozilla::generic_worker
+    include generic_worker::control_bug
 
     case $::operatingsystem {
         Darwin: {
             $taskcluster_host = 'taskcluster'
             $macos_version = regsubst($::macosx_productversion_major, '\.', '')
             $taskcluster_client_id = secret('generic_worker_macosx_client_id')
```
petemoore: Don't we have different client id for staging/production?

dragoscrisan: This change will be part of separate bug

PR #111, opened 2018-07-16
Upgrade most of the non-scriptworker stuff

mozbhearsum:
I haven't done any actual testing, but looking over the changelogs this looks pretty safe. I intend to do some environment pinning for better verification as part of deployment. I looked over the changelogs and the only thing of note was lots of "api-change" entries in boto3 and botocore. The ones I dug into add support for new endpoints, not change existing ones, so I suspect this will all be fine.


escapewindow:
Thanks!

PR #117, opened 2018-07-17
Add pip check to enforce compatible requirements.

Callek:
Lets see if this fixes the false-green

### DIFF
--- a/modules/pushapk_scriptworker/files/requirements.txt
+++ b/modules/pushapk_scriptworker/files/requirements.txt
@@ -46,7 +46,8 @@ oauth2client==4.1.2
 parso==0.3.0
 pexpect==4.6.0
 pickleshare==0.7.4
-prompt_toolkit==2.0.3
+# Prompt toolkit is pinned because of ipython 6.4.0
+prompt_toolkit==1.0.15  # pyup: <2.0.0,>=1.0.15
 ptyprocess==0.6.0
 pyOpenSSL==18.0.0
 pyasn1==0.4.3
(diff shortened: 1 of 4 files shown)


Callek:
With this patch:
```
ipython 6.4.0 has requirement prompt-toolkit<2.0.0,>=1.0.15, but you have prompt-toolkit 2.0.3.
```

modules/pushapk_scriptworker/files/requirements.txt
```diff
@@ -46,7 +46,7 @@ oauth2client==4.1.2
 parso==0.3.0
 pexpect==4.6.0
 pickleshare==0.7.4
-prompt_toolkit==2.0.3
+prompt_toolkit==1.0.15
```
mozbhearsum: You'll need to pin this with the pyup syntax to avoid it being bumped again. Eg: https://github.com/mozilla-releng/build-puppet/blob/master/modules/signingserver/files/linux_requirements.txt#L9

Callek:
We're still failing some py2.7 things with this fixed test, but I'm merging to unbreak pushapk now.

Callek:
> but I'm merging to unbreak pushapk now.

No I'm not after all. (I am fixing pushapk in a seperate patch)

escapewindow:
Pushed the py27 fix, as well as a log parsing fix that puts the errors from pip check at the bottom.

This isn't an ideal fix, since it relies on human-readable output. However, pragmatically it's more usable, we get some sign that it's not working if the grep doesn't work, and we still have the full log if things keep failing after we push the fixes for the parsed errors. We can add parsing for non `pip check` errors, when we hit those errors in the future.

PR #128, opened 2018-07-23
bug 1477434: Bump signcode maxsize to 350MB.

mozbhearsum:
For ASan!


mozbhearsum:
Not sure why Taskcluster isn't showing as done - the linked graph is done and green.

PR #143, opened 2018-07-30
Reverts the slaveapi requirements.txt bug: 1467573

raduiman:


modules/slaveapi/files/requirements.txt
```diff
@@ -44,26 +44,4 @@ boto==2.49.0
 iso8601==0.1.12
 repoze.lru==0.7
 ssh==1.8.0
-urllib3==1.22
```
Akhliskun: We might need this package, as in the file revision before the dashboard broken, it seems we had it.
Revision: [Line47](https://github.com/mozbhearsum/build-puppet/blob/6529f23bbd7850c4c528191ac12a1101dd5667e4/modules/slaveapi/files/requirements.txt#L47)

PR #144, opened 2018-07-30
Update .pyup.yml to run every week on Wednesday

Akhliskun:
After a discussion with @mozbhearsum we have concluded that for the next few weeks we should be doing PyUp merge requests on Thursdays, as Mondays are usually busy with Releases and Merges.

This will stay like this for a few weeks, while CiDuty learns the INs and OUTs about this task, and when the team becomes autonomous this weekly actionable will change either on Saturday or Sunday, when there is a lot less stuff happening. 

@lundjordan If you are okay with the change/plan (review+), Ben will be merging this into the master branch.

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -11,4 +11,4 @@ search: True
 
 # Because updates generally trigger service reloads/restarts, it's better to batch
 # these together when possible.
-schedule: every week
+schedule: every week on thursday
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -11,4 +11,4 @@ search: True
 
 # Because updates generally trigger service reloads/restarts, it's better to batch
 # these together when possible.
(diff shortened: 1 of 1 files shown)


mozbhearsum:
For context, we said Thursday because ciduty can do them before release work starts. Wednesdays are another option, which is generally entirely free of release work. Probably doesn't matter much since we're eventually moving to the weekend.

mozbhearsum:
Ah, actually - I just realized that this should be "on Wednesday" -- because it happens around noon eastern. With this patch, it won't be ready earlier enough on Thursdays.

Akhliskun:
Changed the day to Wednesday as per Ben's comment https://github.com/mozilla-releng/build-puppet/pull/144#issuecomment-408984050

PR #148, opened 2018-08-01
re-enable mdc1 signing servers in signing scriptworkers.

escapewindow:
I'm not bumping signingworker, as that will retire with esr52.


escapewindow:
First test looks good. I think I'd like to see a mac task go green and then I'll be pretty confident.

escapewindow:
Mac looks good.

PR #149, opened 2018-08-02
enable mdc2 prod signing servers, minus mac-v2-signing13

escapewindow:
It looks like the mdc2 servers are ready to go as well, minus [mac-v2-signing13](https://bugzilla.mozilla.org/show_bug.cgi?id=1477139).


escapewindow:
I've got depsigning-worker16 pinned with this password file. Looks good so far... https://tools.taskcluster.net/provisioners/scriptworker-prov-v1/worker-types/depsigning/workers/signing-linux-v1/depsigning-worker16

PR #155, opened 2018-08-03
Disable pip version check to avoid cron spam eating up inboxes.

Callek:
We're getting a bunch of e-mail (every 5 min x 2) about:

> You are using pip version 10.0.1, however version 18.0 is available.
> You should consider upgrading via the 'pip install --upgrade pip' command.

This should stop the bleeding.

For reference we don't manage pip version for the download venv: https://dxr.mozilla.org/build-central/rev/9cb089f5a074bc4d138252914f3d07c8ad6e6229/puppet/modules/puppetmaster/manifests/data.pp#109


mozbhearsum:
+1

PR #157, opened 2018-08-03
Bug 1480412 - fix tc_environment var name

davehouse:
I ended up making the staging workers take production tasks because the "tc_environment" was set to staging, but the in-master module was looking for the variable named "environment".

So this applies the tc_environment to the workerType name set up.


petemoore:
👍

petemoore:
@davehouse For the underlying issue, I've created taskcluster/generic-worker#119...

Apologies for the bustage!

PR #158, opened 2018-08-07
Bug 1480412 - upgrade generic-worker to v10.11.3

davehouse:


modules/packages/manifests/mozilla/generic_worker.pp
```diff
@@ -8,7 +8,11 @@
         'packages::mozilla::generic_worker::end': ;
     }
 
-    $tag = 'v10.8.4'
+    if ($tc_environment == 'staging') {
```
davehouse: I'm tempted to separate these (maybe we move it into the moco-config?) so that we can change staging in-repo instead of pinning to test. And then anyone could submit a pr to test a new version of generic-worker in the beta pool.

petemoore: `tc_environment` feels misleading as a name, especially when soon we'll have an actual staging taskcluster deployment. These are staging workers connected to the production taskcluster environment, rather than connected to a taskcluster staging environment. Perhaps `environment` is a better name than `tc_environment` since it is about the worker being a staging worker, rather than the taskcluster environment being a staging one. I realise you didn't invent the name, but if we're cleaning this up, this might be a good time to do it.

davehouse: :+1:  in releng build-puppet there are aspects for similar attributes, like security levels "high-security" etc and environments "prod" versus "dev". Using an aspect moves the problem to a naming concern for the value ("staging"), but then these workers can be perceived as releng-hardware "staging" and that avoids the conflict with "taskcluster staging".

davehouse: @petemoore how about this? https://github.com/davehouse/build-puppet/pull/1

petemoore: Thanks Dave, I'd say that looks perfect!

dividehex: @petemoore, the `tc_environment` name was encouraged by me because having a node def variable named `environment` was too generic for being in a top-level puppet namespace.  Puppet covers a lot of _other_ infrastructure.  I think @davehouse 's use of aspects is a great solution.

davehouse:
@dividehex thx! we might have prod/dev/staging eventually, or in go-faster :)

modules/packages/manifests/mozilla/generic_worker.pp
petemoore: @dividehex ah thanks for the clarification - that makes perfect sense - I hadn't realised that. ++

PR #159, opened 2018-08-07
enable mdc2 prod signing servers + all mac mdc signing servers

escapewindow:
Currently running on canaries signing-linux-1 and depsigning-worker16.
I'll probably spin mac-v2-signing{9,13} up (now that they're puppetized properly) before landing.


escapewindow:
https://papertrailapp.com/groups/4560112/events?q=mac-v2-signing%20%22%20200%20%22%20mdc looks good so far.

as does https://papertrailapp.com/groups/4560112/events?q=mdc2.mozilla.com%20%22%20200%20%22

PR #160, opened 2018-08-08
Bug 1480512: point puppetize.sh ntpdate at internal ntp host

dividehex:
Turns out the mdc1/2 firewalls are blocking outgoing NTP traffic which is dropping the traffic to pool.ntp.org.  That's ok since we should be using internal NTP servers anyway.   I've redirected the `ntp.build.mozilla.org` cname to the MDC1 infobox host.  This host should be accessible from all internal releng networks.

### DIFF
--- a/modules/puppet/files/puppetize.sh
+++ b/modules/puppet/files/puppetize.sh
@@ -14,6 +14,8 @@
 
 REBOOT_FLAG_FILE="/REBOOT_AFTER_PUPPET"
 OS=`facter operatingsystem`
+NTP_SERVER="ntp.build.mozilla.org"
+
 case "$OS" in
     Darwin) ROOT=/var/root ;;
     *) ROOT=/root ;;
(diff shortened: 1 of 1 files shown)


modules/puppet/files/puppetize.sh
```diff
@@ -102,6 +104,32 @@ function set_osx_names()
...
+        Darwin)
+            if launchctl list org.ntp.ntpd > /dev/null 2>&1 ; then
+                launchctl unload /System/Library/LaunchDaemons/org.ntp.ntpd.plist
+            fi
+            ntpdate $NTP_SERVER
+            launchctl load -w /System/Library/LaunchDaemons/org.ntp.ntpd.plist
+            ;;
+
+        CentOS)
+            ntprunning=`ps ax | grep ntpd | grep -v grep`
```
davehouse: What if we always issue the service stop request and let that fail if it isn't running? Grepping over ps feels like we could make a mistake.

Or, do we have some servers not running ntpd? Maybe we can check service instead of looking for a string in the process list?

dividehex: I agree but it is slightly out of scope and I'm hesitant to change what is already working.

davehouse: +1 and I don't expect anything else to have the "ntpd" string in ps

PR #162, opened 2018-08-09
Bug 1482162: block vnc on mac-v2-signing and macdepsigning

dividehex:
This removes the host-based pf firewall rule allowing incoming vnc connections to mac-v2-signing and depsigning.  Also adds mdc2 to macsigning profile since it was missing. 👎


dividehex:
@escapewindow, yes, the firewall was open on those mdc2 hosts.   The case check on data centers was to help role out the fw rules per DC but I think this got missed since it was set up before mdc2 came online.  I've checked the rest of the profiles and they are fine.  Although, I think at this point we can add a default fail to all the fw profiles in order to catch this in the future.  I'll follow up in another bug/PR.

PR #164, opened 2018-08-13
Fix capitalization of Wednesday for PyPu

Akhliskun:
Based on @mozbhearsum 's findings https://github.com/pyupio/pyup/blob/master/pyup/config.py#L11


mozbhearsum:
It would be nice to send a PR or file an issue for pyup too - this seems like a footgun they may want to fix.

Akhliskun:
Done https://github.com/pyupio/pyup/pull/320

PR #181, opened 2018-08-22
Add missing dependencies.

mozbhearsum:
Not sure why this wasn't caught in tests.


mozbhearsum:
This is to fix issues we're seeing after rolling out the new signingscript, like https://taskcluster-artifacts.net/V7PEDHNaRbaAoA29bXj1Vg/0/public/logs/live_backing.log

PR #200, opened 2018-09-09
Bug 1479619 - add bouncer_check on bm01

nthomas-mozilla:
The plan here is to move bouncer_check from bm81 (scl3) to bm01 (use1). Since there are related nagios checks lets add to bm01 first, move the nagios checks separately, and then bm81 will get cleaned up with the rest of scl3.

### DIFF
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -436,6 +436,7 @@ node 'buildbot-master01.bb.releng.use1.mozilla.com' {
     include toplevel::server::buildmaster::mozilla
     include toplevel::mixin::l10n_bumper
     include toplevel::mixin::releaserunner3
+    include toplevel::mixin::bouncer_check
 }
 
 node 'buildbot-master02.bb.releng.use1.mozilla.com' {


nthomas-mozilla:
Please go ahead with merging if it looks OK.

MihaiTabara:
Related news, nice to see bouncer check running on beta after trains were merged the other week. https://treeherder.mozilla.org/#/jobs?repo=mozilla-beta&selectedJob=198254771

nthomas-mozilla:
Sorry for the missing context. hwine wrote this check originally as a python package, the guts of which is [check_bouncer.py in hg:build/nagios-tools](https://hg.mozilla.org/build/nagios-tools/file/tip/nagios_tools/scripts/check_bouncer.py). You can see what it checks at line 65 in `default_products`, so it covers aliases for nightly/deved/beta/release/esr.  The checks run every 5 minutes in nagios.

In turn I'm not familiar with the new checks, but from the log it looks like it concentrates on the versioned products and runs at the end of each release, so they are complementary.

Thanks for the review & merge.

MihaiTabara:
Ah, good to know! more tests is better! CC @srfraser

PR #205, opened 2018-09-11
Bug 1490326 - releaserunner: disable betas

rail:
* Get rid of devedition_pattern completely
* firefox_pattern should only accept releases and ESR
* remove releases/mozilla-beta from allowed_branches

### DIFF
--- a/manifests/moco-config.pp
+++ b/manifests/moco-config.pp
@@ -310,9 +310,9 @@ class config inherits config::base {
             partner_min_version                   => 61,
             releaserunner_config_file             => 'release-runner.yml',
             allowed_branches                      => [ 'projects/jamun', 'projects/maple', 'projects/birch' ],
-            fennec_pattern                        => 'Fennec-.*',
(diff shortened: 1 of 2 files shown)


manifests/moco-config.pp
```diff
@@ -310,9 +310,9 @@
             partner_min_version                   => 61,
             releaserunner_config_file             => 'release-runner.yml',
             allowed_branches                      => [ 'projects/jamun', 'projects/maple', 'projects/birch' ],
-            fennec_pattern                        => 'Fennec-.*',
-            firefox_pattern                       => 'Firefox-(5[89]|6[0-9]).*',
-            devedition_pattern                    => 'Devedition-(5[89]|6[0-9]).*',
+            # Exclude betas
+            fennec_pattern                        => 'Fennec-\d+\.\d+(\.\d+)?-.*',
+            firefox_pattern                       => 'Firefox-\d+\.\d+(\.\d+)?(esr)?-.*',
```
tomprince: I think we should drop the fennec/firefox changes in this hunk, to make using ship-it-v1 on maple easier.

PR #209, opened 2018-09-12
Switch to autograph prod.

mozbhearsum:
I think we want to switch to prod everywhere, including depsigning? I'll need to add some entries to Hiera before this lands.


escapewindow:
> I think we want to switch to prod everywhere, including depsigning?

I *think* so. Then we can just make sure prod stays up, and we have more leeway to test things or bounce stage without breaking integration trees.

> I'll need to add some entries to Hiera before this lands.
+1

PR #210, opened 2018-09-13
Revert "scriptworker 16.0.0"

rail:
Reverts mozilla-releng/build-puppet#208

From IRC:
> tomprince: Scriptworker 16.0 is going to cause release tasks to retry endlessly (it doesn't handle that relpro actions aren't hooks) so we should probably back it out before starting any releases.
> ...
> tomprince: Just need to revert the build-puppet commit that upgraded things.
> rail: revert https://github.com/mozilla-releng/build-puppet/pull/208?
> tomprince: Yeah.


rail:
Just playing safe...

rail:
@escapewindow JFYI

escapewindow:
ah, boo

PR #213, opened 2018-09-14
Remove scl3 references

dividehex:
This is PR is a **BIG** tear out of all things referencing scl3 since the scl3 datacenter is just about completely offline at this time.  Once this patch has landed, we can follow up and clean out modules that aren't in use any more such as slaveapi, awsmanagers, and builtbot.


manifests/moco-config.pp
```diff
@@ -168,16 +151,12 @@
     $signing_allowed_ips             = [
         '10.134.68.32/32', # dev-master2
         '10.26.48.41/32',  # partner-repack1
-        '10.26.44.0/22',   # wintry.releng.scl3 (vlan 244)
-        '10.26.52.0/22',   # build.releng.scl3 (vlan 252)
-        '10.26.64.0/22',   # try.releng.scl3 (vlan 264)
```
nthomas-mozilla: We can reduce this and `signing_new_token_allowed_ips` even further but I'll make a separate pull for that.

mozbhearsum:
FYI: it looks like the AWS log aggregators may be actively broken, and will be fixed by this. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1491454 for that last week. (Could be wrong, I didn't look at it in great detail.)

dividehex:
> 
> 
> FYI: it looks like the AWS log aggregators may be actively broken, and will be fixed by this. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1491454 for that last week. (Could be wrong, I didn't look at it in great detail.)

This actually doesn't fix collectd for aws hosts since there are no attributes or defaults defined for aws hosts. I wonder how it was ever configured to write to scl3... what am I missing?

manifests/moco-config.pp
dividehex: 👍

PR #216, opened 2018-09-17
Fix fw

dividehex:


dividehex:
This fixes the fallout from https://github.com/mozilla-releng/build-puppet/pull/213.  The slaveapi was still being referenced on osx taskcluster workers plus a couple other things.

PR #218, opened 2018-09-18
Bug 1491726 - remove old IP blocks for signing

nthomas-mozilla:
I did a bunch of grepping to make sure nothing is using the removed definitions.


modules/fw/manifests/networks.pp
```diff
@@ -36,18 +36,10 @@
     # AWS Networks; See https://github.com/mozilla-releng/build-cloud-tools/blob/master/configs/subnets.yml
 
     # USE1 Network CIDRs
```
escapewindow: We may want to verify that these are buildbot-worker-only subnets... i.e., no taskcluster workers.

modules/fw/manifests/roles/dep_signing_from_osx.pp
```diff
@@ -8,10 +8,6 @@
     fw::rules { 'allow_dep_signing_from_osx':
         # All signing sources
         sources =>  [   $::fw::networks::all_partner_repack,
```
escapewindow: do we need `all_partner_repack` anymore? in this file and the other `*signing_from_*` files.

modules/fw/manifests/networks.pp
dividehex: If these subnets are verified not be in use anymore, we should make sure to delete them from the vpc

modules/fw/manifests/networks.pp
```diff
@@ -36,18 +36,10 @@
...
-    $use1_test    = [ '10.134.40.0/21', '10.134.56.0/22', '10.134.60.0/22', '10.134.156.0/22' ]
-    $use1_build   = [ '10.134.52.0/22', '10.134.164.0/24' ]
-    $use1_srv     = [ '10.134.48.0/22' ]
```
dividehex: srv has typically been were we put infra like puppet masters and log aggregators.  I assume it should stick around.

modules/fw/manifests/networks.pp
nthomas-mozilla: aki, I've checked and they don't overlap. In use1, signing scriptworkers are all in 10.134.30.0/24, and non-signing scriptworkers in 10.134.48.0/24; in usw2, all scriptworkers are in 10.132.30.0/24.

dividehex, and update cloud-tools I guess. More than I was planning on right now, I'll see if I can get to it.

modules/fw/manifests/networks.pp
nthomas-mozilla: Agreed, and we put non-signing scriptworkers in there too for use1, but this definition doesn't seem to be used anywhere else in puppet. Would you prefer to leave it as a reference ?

modules/fw/manifests/roles/dep_signing_from_osx.pp
nthomas-mozilla: Nope, I'll remove it.

modules/fw/manifests/networks.pp
dividehex: Yes.  I'd prefer to keep the reference even if it isn't used.

PR #222, opened 2018-09-19
Bug 1488396 - Update pushsnap_scriptworker to latest snapcraft

JohanLorenzo:
I rebuilt the dependency list with https://github.com/mozilla-releng/pushsnapscript/pull/13. Maybe bumping some of the deps caused [bug 1488396](https://bugzil.la/1488396). I'm looking into it.

### DIFF
--- a/modules/pushsnap_scriptworker/files/requirements.txt
+++ b/modules/pushsnap_scriptworker/files/requirements.txt
@@ -16,7 +16,7 @@ idna==2.7
 idna_ssl==1.1.0
 json-e==2.7.0
 jsonschema==2.6.0
-# Puppet check is deactivated because it requires some distro library to be installed.
+# Puppet check is deactivated because it requires some distro library (libsodium) to be installed.
 libnacl==1.3.6    # puppet: nodownload
 mohawk==0.3.4
 multidict==4.4.2
@@ -24,7 +24,8 @@ pexpect==4.6.0
 progressbar33==2.4
 ptyprocess==0.6.0
 pyelftools==0.24
-pymacaroons==0.9.2
+# Puppet check is deactivated because it requires some distro library (libsodium) to be installed.
+pymacaroons==0.9.2    # puppet: nodownload
 python-dateutil==2.7.3
 python-debian==0.1.28
 python-gnupg==0.4.3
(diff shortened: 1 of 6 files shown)


modules/pushsnap_scriptworker/files/requirements.txt
```diff
@@ -16,29 +13,28 @@ idna==2.7
 idna_ssl==1.1.0
 json-e==2.7.0
 jsonschema==2.6.0
+libnacl==1.3.6
```
escapewindow: We probably need to add a ` # puppet: nodownload` comment here -- libnacl downloads seem to barf due to required system libs that aren't on puppet masters.

tomprince: It looks like `libnacl` is a dependency of `pymacaroons`. Newer versions of that use `pynacl`, which has wheels available. If we could upgrade to that, it would avoid the trouble here.

JohanLorenzo: Thanks for the context! I would prefer to avoid updating until the push-snap task gets back to green. pushsnapscript has a very dirty hack where we only import a part of the dependencies of snapcraft. The reasons of this hack are:

* snapcraft has some Ubuntu-specific deps and we're running it on CentOS
* there is no snapcraft package on CentOS 6.
* we can't run the snap package of snapcraft because CentOS 6 is not supported.

If I manage to get the next job, I plan to comment `requirements.txt` to explain why some deps aren't up to date. What do you guys think?

escapewindow: Sure. However, not having the `# puppet: nodownload` is causing us to get puppet mail multiple times an hour, with the error message `Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-download-34e2ruc8/libnacl/`. Could you update your environment with that comment, at least, to reduce the mail errors?

JohanLorenzo: Oh, I'm sorry for the breakage. I didn't know personal env were subject to this kind of checks. This was fixed 5 hours ago.

JohanLorenzo:
Latest revision enabled https://tools.taskcluster.net/groups/CTU6CA-PTGyDBu4IExIgoA/tasks/USJ2lRdtQ16waOO0kzT-zw/runs/2/logs/public%2Flogs%2Flive_backing.log.

Please note xdelta3 is now mandatory because https://bugzilla.mozilla.org/show_bug.cgi?id=1493053#c0

PR #223, opened 2018-09-19
Use starttls when connecting to ldap

dividehex:


dividehex:
This was requested by IT.  See: https://bugzilla.mozilla.org/show_bug.cgi?id=1492157

PR #225, opened 2018-09-20
Kill old modules

mozbhearsum:
This removes a bunch of what appear to be unused modules, toplevel manifests, and node definitions:
* All buildbot master nodes except 01/77, and the buildbot bits from them.
* "aws" module (used by build machines in AWS, I think)
* aws_manager, buildbot_bridge*, buildmaster, cruncher, runner, signingworker, slaveapi, slaverebooter modules + associated toplevel manifests
* "slave" toplevel manifests (which appear to be long dead), and a few other misc. ones like selfserve_agent.


nthomas-mozilla:
There's also a slaveapi blacklist entry in `.pyup.yml` which could go too.

PR #227, opened 2018-09-21
Bug 1493265 Update datadog-statsd port r=aki

srfraser:


escapewindow:
We probably want a followup here to move this port into `settings.pp`; we can then also populate the `script_config.json` template with that setting.

A followup is ok here, but when running this and https://github.com/mozilla-releng/signingscript/pull/73 together in my staging env, I got this error:

```
2018-09-21 20:48:30,940 - datadog.api - INFO - No agent or invalid configuration file found
```

The error persisted after a reboot. My puppet env is using [this branch](https://github.com/mozilla-releng/build-puppet/compare/master...escapewindow:staging?expand=1).

I'm going to leave this for more debugging next week. Thanks for jumping on this today!

escapewindow:
`/etc/dd-agent/datadog.conf` appears to have the updated port in it...

srfraser:
Setting this up locally seems to work, so I'm getting a loaner to examine what puppet's doing that might be different.

The error is raised at https://github.com/DataDog/datadogpy/blob/master/datadog/util/hostname.py#L59  so it's likely something in https://github.com/DataDog/datadogpy/blob/master/datadog/util/config.py#L105 having trouble.

srfraser:
Using a puppetised instance I don't get this error - can you share the datadog.conf that's being used, or the instance name?

escapewindow:
I believe this was on signing-linux-dev1.

PR #229, opened 2018-09-24
 Bug 1474570 - Linux workers on moonshot in mdc2 run generic-worker

dragoscrisan:
Linux workers on moonshot in mdc2 run generic-worker

### DIFF
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -12,7 +12,11 @@ node /^t-yosemite-r7-\d+\.test\.releng\.(mdc1|mdc2)\.mozilla\.com$/ {
     include toplevel::worker::releng::generic_worker::test::gpu
 }
 
-# Linux on moonshot in mdc1 running taskcluster worker
+# Linux on moonshot in mdc1 running taskcluster-worker, but will be migrated to generic-worker once bug 1474570 lands
+# The migration is underway such that all taskcluster-worker workload is being moved from worker type
+# gecko-t-linux-talos to gecko-t-linux-talos-tw, and that when that completes, gecko-t-linux-talos will then be used
+# for generic-worker implementation of talos linux tasks. For more details, please see:
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1474570#c32
 node /^t-linux64-(ms|xe)-\d{3}\.test\.releng\.mdc1\.mozilla\.com$/ {
     $aspects          = [ 'low-security' ]
     $slave_trustlevel = 'try'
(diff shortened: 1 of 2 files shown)


manifests/moco-nodes.pp
```diff
@@ -21,18 +25,13 @@
...
-# The migration is underway such that all taskcluster-worker workload is being moved from worker type
-# gecko-t-linux-talos to gecko-t-linux-talos-tw, and that when that completes, gecko-t-linux-talos will then be used
-# for generic-worker implementation of talos linux tasks. For more details, please see:
-# https://bugzilla.mozilla.org/show_bug.cgi?id=1474570#c32
+# Linux on moonshot in mdc2 running generic-worker
 
 node /^t-linux64-(ms|xe)-\d{3}\.test\.releng\.mdc2\.mozilla\.com$/ {
     $aspects          = [ 'low-security' ]
     $slave_trustlevel = 'try'
-    $taskcluster_worker_type  = 'gecko-t-linux-talos'
```
petemoore: I see that the worker type is currently determined from [this line](https://github.com/mozilla-releng/build-puppet/blob/2a898a8f181bcebcc8896c14bbae8f2741aaf627/modules/generic_worker/manifests/init.pp#L95) - however, that worker type is incorrect (linux workers shouldn't have a worker type containing `osx`).

On reflection, I think it would be better to explicitly define the worker type names in the node declaration here in `moco-nodes.pp` rather than reverse engineer it in `modules/generic_worker/manifests/init.pp`. This method worked ok when e.g. all production macs had the same worker type name, but doesn't scale well when we support multiple worker types in production. At the moment, we only have linux talos for generic-worker, but when docker support is added to generic-worker, we'll be running potentially multiple worker types per platform, so maybe this is a good time to migrate the worker type settings from `modules/generic_worker/manifests/init.pp` to `manifests/moco-nodes.pp`. This also to my mind improves clarity and removes some of the magic.

petemoore: @dragoscrisan I think this still needs addressing:

> On reflection, I think it would be better to explicitly define the worker type names in the node declaration here in `manifests/moco-nodes.pp` rather than reverse engineer it in `modules/generic_worker/manifests/init.pp`. This method worked ok when e.g. all production macs had the same worker type name, but doesn't scale well when we support multiple worker types in production. At the moment, we only run generic-worker on linux in the case of linux talos tests; however when docker support is added to generic-worker, we'll be running potentially multiple worker types per platform, so maybe this is a good time to migrate the worker type settings from `modules/generic_worker/manifests/init.pp` to `manifests/moco-nodes.pp`. This also (to my mind) improves clarity (it is a bit more explicit) and reduces the magic.

petemoore:
@dragoscrisan I think this still needs addressing:

> On reflection, I think it would be better to explicitly define the worker type names in the node declaration here in `manifests/moco-nodes.pp` rather than reverse engineer it in `modules/generic_worker/manifests/init.pp`. This method worked ok when e.g. all production macs had the same worker type name, but doesn't scale well when we support multiple worker types in production. At the moment, we only run generic-worker on linux in the case of linux talos tests; however when docker support is added to generic-worker, we'll be running potentially multiple worker types per platform, so maybe this is a good time to migrate the worker type settings from `modules/generic_worker/manifests/init.pp` to `manifests/moco-nodes.pp`. This also (to my mind) improves clarity (it is a bit more explicit) and reduces the magic.

manifests/moco-nodes.pp
dividehex: I agree with you Pete, it's a lot easier to track down the work type if you simply look at the node def. 👍

dragoscrisan:
Using PR https://github.com/mozilla-releng/build-puppet/pull/245 for generic-worker refactoring:

PR #231, opened 2018-09-25
Bug 1491335 - enable beetmoverworkers handle android-components.

MihaiTabara:


manifests/moco-nodes.pp
```diff
@@ -383,6 +383,8 @@
     $beetmoverworker_env = 'mobile-staging'
     $timezone                 = 'UTC'
     $only_user_ssh            = true
+    $pin_puppet_server = 'releng-puppet2.srv.releng.mdc1.mozilla.com'
```
MihaiTabara: For the record, this is temporary and will go away before merging. I just use this as it's easier to edit on my local machine and pull in my puppet environment.

modules/beetmover_scriptworker/files/requirements.txt
```diff
@@ -6,7 +6,7 @@ aiohttp==3.4.4
 arrow==0.12.1
 async_timeout==3.0.0
 attrs==18.2.0
-beetmoverscript==7.9.1  # puppet: nodownload
+beetmoverscript==8.0.0.dev0  # puppet: nodownload
```
MihaiTabara: For the record, this is temporary and will go away before merging. I just use this as it's easier to edit on my local machine and pull in my puppet environment.

manifests/moco-nodes.pp
JohanLorenzo: Nit: Remove this line and the following one.

modules/beetmover_scriptworker/files/requirements.txt
JohanLorenzo: Nit: 8.0.0

modules/beetmover_scriptworker/manifests/settings.pp
```diff
@@ -148,24 +150,21 @@
             verify_cot_signature                    => true,
             cot_product                             => 'thunderbird',
         },
```
MihaiTabara: Should we define the `mobile-staging` too since we're here and we have new credentials?

modules/beetmover_scriptworker/manifests/settings.pp
```diff
@@ -30,6 +30,7 @@
                 firefox     => 'mozilla-releng-dep-partner',
             },
 
+            # TODO rename secret to dep_maven_geckoview_beetmover_aws_*
```
MihaiTabara: 👍

modules/beetmover_scriptworker/manifests/settings.pp
```diff
@@ -148,24 +150,21 @@
...
             taskcluster_access_token                => secret('beetmoverworker_prod_taskcluster_access_token_mobile'),
-            taskcluster_scope_prefix                => 'project:mobile:android-components:releng',
-            # TODO to turn this back on
-            sign_chain_of_trust                     => false,
+            taskcluster_scope_prefix                => 'project:mobile:android-components:releng:beetmover:',
+            sign_chain_of_trust                     => true,
             verify_chain_of_trust                   => true,
-            # TODO to turn this back on
-            verify_cot_signature                    => false,
+            verify_cot_signature                    => true,
```
MihaiTabara: Even with these turned on, things should work smoothly, because we don't yet have the `json-e` right? For some reason I'm still worried a bit as I haven't added cot-gpg keys for this particular instance and I'm unsure if this will break it or not.

modules/beetmover_scriptworker/manifests/settings.pp
JohanLorenzo: Let's do it in a follow up. We don't have all creds ready yet

modules/beetmover_scriptworker/manifests/settings.pp
JohanLorenzo: That's right, it will. json-e is skipped at this stage https://github.com/mozilla-releng/scriptworker/blob/d7ac2ee55e7fb537b939090358692ebd2131b725/scriptworker/cot/verify.py#L1218.

There is no need to create a new cot-gpg key these days. We only rely on a single key https://github.com/mozilla-releng/cot-gpg-keys/tree/a53350f58c91d3bd9845bf6ff123f588ec2e43c4/scriptworker/valid

MihaiTabara: 👍

PR #236, opened 2018-09-26
Lzma python

escapewindow:
This patch allows for a python 3.6.5-2 with lzma support. (That, in turn, unblocks the use of mardor in signingscript.)

I've tested this: signingscript-with-hash-signing works when testing https://github.com/mozilla-releng/signingscript/pull/69 using this custom build of python. I also tested against the current signingscript, after I uninstalled `xz-devel`; that appears to still work, though I haven't tested on a pristine environment (installing and uninstalling xz-devel may leave some cruft?). The main downside of this patch is that the scriptworkers delete their virtualenvs when this lands, leading to lengthy downtimes. We'll want to land when the tree is fairly quiet.

### DIFF
--- a/modules/packages/manifests/mozilla/python3.pp
+++ b/modules/packages/manifests/mozilla/python3.pp
@@ -25,7 +25,7 @@ class packages::mozilla::python3 {
             Anchor['packages::mozilla::python3::begin'] ->
             package {
                 'mozilla-python36':
-                    ensure => '3.6.5-1.el6';
+                    ensure => '3.6.5-2.el6';
             } -> Anchor['packages::mozilla::python3::end']
         }
         Darwin: {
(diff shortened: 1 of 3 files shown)


escapewindow:
I didn't time it, but it was not a fast thing. 30min to an hour plus? I debated getting a staging env up that pins everything, and then pointing all the scriptworkers at it piecemeal for a manual rollout. Once everything's updated we could roll out to prod and then unpin everything.

escapewindow:
Good news: I installed lzma-python 3.6.5 on depsigning-worker1, and it's taking tasks and going green.

Bad news: the upgrade takes ~70min. But things are pretty quiet, and I can upgrade a percentage of the pool(s) via staging puppet envs.

PR #240, opened 2018-09-27
Bug 1489603: restart mig if exists

davehouse:
Add "only-if" check for mig-agent restart to check for the service before trying to restart it.

* The mig-agent.service is not created until mig-agent has run (not set up in the installation package but by the binary itself).
* Systemctl restart mig-agent.service will return a failure exitcode and Puppet will then return a failure exist code.
Puppetize.sh will keep running puppet until it gets a truthy/0 exit, but mig-agent will never have been run and so the problem repeats.

### DIFF
--- a/modules/mig/manifests/agent/daemon.pp
+++ b/modules/mig/manifests/agent/daemon.pp
@@ -19,10 +19,14 @@ class mig::agent::daemon {
             case $::operatingsystemrelease {
                 16.04: {
                     exec {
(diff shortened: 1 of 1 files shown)


davehouse:
We 've run into issues with mig-agent failing (appears to be because of an external ip resolution service?), and so switching to this start/stop of it (through systemd) prevents that failure from stopping our puppet run.

PR #242, opened 2018-09-28
mobile: use autograph-external domain

milescrabill:
this is a no-op as right now, autograph-external points to autograph. we will be switching the autograph DNS over to an internal only LB in the future, hence this change.


escapewindow:
@JohanLorenzo - right now autograph-external and autograph are CNAMEs, so I think this is ok. Just wanted to pass it by you first.

PR #243, opened 2018-09-28
Move a number of linux mdc2 workers from gecko-t-linux-talos to gecko…

dragoscrisan:
Move a number of linux mdc2 workers from gecko-t-linux-talos to gecko-t-linux-talos-tw

### DIFF
--- a/manifests/moco-nodes.pp
+++ b/manifests/moco-nodes.pp
@@ -40,7 +40,7 @@ node /^t-linux64-(ms|xe)-[3-4][0-9][0-9]\.test\.releng\.mdc2\.mozilla\.com$/ {
 node /^t-linux64-(ms|xe)-[5][0-9][0-9]\.test\.releng\.mdc2\.mozilla\.com$/ {
     $aspects          = [ 'low-security' ]
     $slave_trustlevel = 'try'
-    $taskcluster_worker_type  = 'gecko-t-linux-talos-tw'
+    $taskcluster_worker_type  = 'gecko-t-linux-talos'
     include fw::profiles::linux_taskcluster_worker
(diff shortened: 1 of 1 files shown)


manifests/moco-nodes.pp
```diff
@@ -21,13 +21,23 @@
     include toplevel::worker::releng::taskcluster_worker::test::gpu
 }
 
+# Move some workers from mdc2 into gecko-t-linux-talos-tw queue
```
petemoore: nit: `[3-4]` can be written more simply as `[34]`

Can you add a comment that explains which hosts exist, and what they are assigned to, so it is possible for an observer to see how many machines are in each pool, and how they are named?

The problem with the change as it is now is that it isn't possible to deduce the pool sizes. With a comment, everyone can see this. I've just made up numbers below, but my idea would be something like this:

```
# The pool of xe/moonshot hosts is divided into worker types as follows:
#
# gecko-t-talos-linux: 
# mdc1
#   134 moonshot hosts:  t-linux64-ms-{353..486}
#   25 xe hosts:         t-linux64-xe-{487..511}
# mdc2
#   134 moonshot hosts:  t-linux64-ms-{353..486}
#   25 xe hosts:         t-linux64-xe-{487..511}
# total:
#   318 hosts
#
# gecko-t-talos-linux-tw: 
# mdc1
#   134 moonshot hosts:  t-linux64-ms-{353..486}
#   25 xe hosts:         t-linux64-xe-{487..511}
# mdc2
#   134 moonshot hosts:  t-linux64-ms-{353..486}
#   25 xe hosts:         t-linux64-xe-{487..511}
# total:
#   318 hosts
```

PR #255, opened 2018-10-08
Autograph hash

escapewindow:
Add autograph hash-signing support. This patch:

- adds `mar` as a python dependency.
  - also pins `datadog` to 0.17.0 since it's pinned in signingscript. NI'ed @srfraser [to see if we want to keep that pinned](https://github.com/mozilla-releng/signingscript/commit/febe8842da6b5ca00d7fe332caa37bc4f1473925#r30817586).
- adds `xz-devel` as a signing scriptworker package
- bumps us to signingscript 9.2.0, which adds autograph-hash support
- adds the `autograph_hash_only_mar384` format to the passwords templates.

### DIFF
--- a/modules/signing_scriptworker/files/requirements.txt
+++ b/modules/signing_scriptworker/files/requirements.txt
@@ -2,34 +2,42 @@
 PyYAML==3.13
 aiohttp==3.4.4
 arrow==0.12.1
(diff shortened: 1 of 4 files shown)


escapewindow:
@nthomas-mozilla , is this something you're comfortable reviewing?

PR #258, opened 2018-10-09
Update hg server cert for bug 1495464

nthomas-mozilla:
Copy and paste from [bug 1495464 comment #6](https://bugzilla.mozilla.org/show_bug.cgi?id=1495464#c6). I'm pretty sure that mercurial.ini is unused now that we don't run any Windows builders but update it to the v3.9+ style config too. 

Also removes ftp-ssl.mozilla.org, which is baffling in an hg config file.


nthomas-mozilla:
I've run my eye down moco-nodes.pp and there aren't any Windows hosts left. We can probably remove that support at some point.  It turns out ftp-ssl.m.o was from when we used to host bundles there, prior to hg.m.o using s3.

PR #259, opened 2018-10-10
Add nagios check for roller server

dragoscrisan:
Add nagios check for roller server


modules/packages/manifests/openssl.pp
```diff
@@ -48,8 +48,23 @@
                 }
                 16.04: {
                     package {
-                        ['openssl', 'libssl1.0.0', 'libssl-dev']:
+                        ['openssl', 'libssl-dev']:
+                            ensure => '1.0.2g-1ubuntu4.6';
+                    }
+                    if $::fqdn =~ /^roller/ {
+                        realize(Packages::Aptrepo['openssl'])
```
dividehex: This shouldn't be limited to the roller hosts only.  If nrpe requires an updated set of openssl packages then all ubuntu 16.04 hosts should get that update.

PR #266, opened 2018-10-16
Bug 1492245 - part 1: Allow Focus signing via autograph

JohanLorenzo:
After making this patch, I realized I need to update the decision task in Focus too. Signingscript know whether to contact autograph only if the format starts with `autograph_`: https://github.com/mozilla-releng/signingscript/blob/eb321067005541f3fcdf65e9375aef7f4f20ff5f/signingscript/utils.py#L169

### DIFF
--- a/modules/signing_scriptworker/templates/passwords-mobile.json.erb
+++ b/modules/signing_scriptworker/templates/passwords-mobile.json.erb
@@ -5,7 +5,8 @@
         ["signing9.srv.releng.mdc1.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"],
         ["signing10.srv.releng.mdc2.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"],
         ["signing11.srv.releng.mdc2.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"],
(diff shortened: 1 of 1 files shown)


modules/signing_scriptworker/templates/passwords-mobile.json.erb
```diff
@@ -1,11 +1,6 @@
...
-        ["signing9.srv.releng.mdc1.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"],
-        ["signing10.srv.releng.mdc2.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"],
-        ["signing11.srv.releng.mdc2.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"],
-        ["signing12.srv.releng.mdc2.mozilla.com:9120", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_release_password"]) %>", ["focus-jar"], "signing_server"]
+        ["http://autograph-external.prod.autograph.services.mozaws.net/", "<%= scope.function_secret(["autograph_focus_prod_username"]) %>", "<%= scope.function_secret(["autograph_focus_prod_password"]) %>", ["autograph_focus"], "autograph"]
```
MihaiTabara: @JohanLorenzo I wonder if it's `autograph_apk` instead of `autograph_focus` ?

MihaiTabara: Nevermind, these are all values consumed by scriptworker, hence invisible to Autograph, so they can be whatever we want them to be.

JohanLorenzo: Yup, signingscript is the sole consumer of that value.

JohanLorenzo:
Discussed over IRC. Let's allow both signing format. This enables (for instance) Nightlies to be signed with Autograph and releases with the signing servers.

PR #269, opened 2018-10-17
Bug 1492245 - Swap Fennec dep-signing to Autograph

MihaiTabara:


modules/signing_scriptworker/templates/passwords.json.erb
```diff
@@ -20,27 +20,6 @@
         ["mac-v2-signing13.srv.releng.mdc2.mozilla.com:9100", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_nightly_password"]) %>", ["macapp"], "signing_server"],
         ["https://autograph-external.prod.autograph.services.mozaws.net", "<%= scope.function_secret(["autograph_mar_nightly_username"]) %>", "<%= scope.function_secret(["autograph_mar_nightly_password"]) %>", ["autograph_mar384", "autograph_hash_only_mar384"], "autograph"]
     ],
```
MihaiTabara: The way I understood this is that this is duplicate definition with the one from `dep-passwords.json.erb` and the reason we have it is that we might want occasionally to still be able to sign on the production signing-servers with the `dep` key. If my understanding is right, this is not needed right now and we could clean this up.

Currently, machines using this template file are:
* [signing-linux-v1](https://tools.taskcluster.net/provisioners/scriptworker-prov-v1/worker-types/signing-linux-v1) from [here](https://github.com/mozilla-releng/build-puppet/blob/master/manifests/moco-nodes.pp#L275)
* [tb-signing-v1](https://tools.taskcluster.net/provisioners/scriptworker-prov-v1/worker-types/tb-signing-v1) from [here](https://github.com/mozilla-releng/build-puppet/blob/master/manifests/moco-nodes.pp#L299)

None of which, AFAIK, are using the `dep` definition whatsoever. 
CC'ing @escapewindow and @tomprince

modules/signing_scriptworker/templates/dep-passwords.json.erb
```diff
@@ -18,6 +18,7 @@
         ["mac-v2-signing11.srv.releng.mdc2.mozilla.com:9110", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_dep_password"]) %>", ["macapp"], "signing_server"],
         ["mac-v2-signing12.srv.releng.mdc2.mozilla.com:9110", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_dep_password"]) %>", ["macapp"], "signing_server"],
         ["mac-v2-signing13.srv.releng.mdc2.mozilla.com:9110", "<%= scope.function_secret(["signing_server_username"]) %>", "<%= scope.function_secret(["signing_server_dep_password"]) %>", ["macapp"], "signing_server"],
+        ["https://autograph-2.stage.mozaws.net", "<%= scope.function_secret(["autograph_fennec_dep_username"]) %>", "<%= scope.function_secret(["autograph_fennec_dep_password"]) %>", ["autograph_fennec"], "autograph"]
```
MihaiTabara: @JohanLorenzo Am a bit confused as to why I had to use the staging instance. Am looking below and I see the `mar` `dep` counterpart is talking to the production instance. Should we leave this be as is?

MihaiTabara:
Leftovers, if my understanding is correct is:
a) land in-tree patch to maple to replace `jar` with `autograph_fennec` in the formatting scopes
b) land this PR to affect all dep signing workers
c) confirm signing is working properly - Fennec jars should look good
d) land to `mozilla-central`

modules/signing_scriptworker/templates/passwords.json.erb
JohanLorenzo: /me defers to Aki and Tom

modules/signing_scriptworker/templates/dep-passwords.json.erb
JohanLorenzo: That's a great question. I'm not too sure what's the scope of "stage" in `autograph-2.stage`. Is it meant to host non-production keys? Or is it meant to test autograph changes? Or both, @g-k?

modules/signing_scriptworker/templates/dep-passwords.json.erb
```diff
@@ -18,6 +18,7 @@
...
         ["https://autograph-external.prod.autograph.services.mozaws.net", "<%= scope.function_secret(["autograph_mar_dep_username"]) %>", "<%= scope.function_secret(["autograph_mar_dep_password"]) %>", ["autograph_mar384", "autograph_hash_only_mar384"], "autograph"]
```
MihaiTabara: @g-k I'm confused a bit, please bear with me. You mentioned that the `mar dep` key is on the staging, yet here it's defined to access the production instance (line 22 in `dep-passwords.json.erb`) so if I understand this right, right now the dep-mar signing-scriptworkers will talk to production instance of Autograph. If that's the case, I see two possible scenarios:

a) we standardize that all dep keys should lie within the staging instance of Autograph, in which case we'll have to move the current `mar dep` key too
b) we standardize that all dep keys lie within the production instance of Autograph, in which case we'll have to move current `fennec dep` key there too

Does this make sense or have I misunderstood something?

g-k: @MihaiTabara yeah, my mistake. I just checked the config and the MAR dep key is in prod.

MihaiTabara: @g-k no problem, thanks for clarifications! Since we already have `mar dep` on production Autograph, let's keep that as a standard. So let's put all dep keys in the production Autograph instance and not touch staging. 

Do you need to generate another set of credentials or can you transplant those to prod?

modules/signing_scriptworker/templates/passwords.json.erb
escapewindow: I believe so. I don't think it's harmful to keep it, but I also don't think it'll break anything to clean it up.

modules/signing_scriptworker/templates/dep-passwords.json.erb
escapewindow: My thinking was, we often want to run staging releases, and we use dep keys in integration (though not for mar signing). Since I assume autograph-stage doesn't have the same expectations around uptime as autograph-prod, let's use autograph-prod for dep signing in production.

MihaiTabara: Switched back to prod instance here. Once the Autograph deployment moves the configs from staging to production, we're green light to test here.

MihaiTabara: @g-k deployed this today, proceeding with merging this and testing the dep-signing Fennec.

modules/signing_scriptworker/templates/passwords.json.erb
MihaiTabara: Okay, will proceed with the cleaning. I'll keep an eye on things. Thanks!

PR #272, opened 2018-10-19
bug 1499809 - l10n bumper should touch a file on success.

escapewindow:
update the gecko revision to the landing of the initial patch from bug 1499809.


escapewindow:
Me too, then I went to https://hg.mozilla.org/mozilla-central/rev/d0d593bf9772 and didn't see the long hash anywhere, and didn't dig.

PR #279, opened 2018-10-29
Osslsigncode bump

Callek:
Ignore the pinning patch of this (for now) but if this looks good I'll deploy.

### DIFF
--- a/modules/packages/manifests/mozilla/osslsigncode.pp
+++ b/modules/packages/manifests/mozilla/osslsigncode.pp
@@ -8,7 +8,7 @@ class packages::mozilla::osslsigncode {
             realize(Packages::Yumrepo['osslsigncode'])
             package {
                 'osslsigncode':
-                    ensure => "1.7.1-1.el6";
+                    ensure => "1.7.1-2.el6";
             }
         }
 
--- a/modules/packages/manifests/mozilla/osslsigncode.spec
+++ b/modules/packages/manifests/mozilla/osslsigncode.spec
@@ -1,7 +1,7 @@
 Summary: Tool for Authenticode signing of EXE/CAB files
 Name: osslsigncode
 Version: 1.7.1
-Release: 1%{?dist}
+Release: 2%{?dist}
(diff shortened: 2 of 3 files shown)


escapewindow:
Have you tested this?

I avoid landing the test patches to master by:

1. writing the patches (done)
1. pushing to github, e.g. github/escapewindow/build-puppet, in the staging-aki branch.
1. pinning the right machine in the applicable staging branch
1. update my environment to the right branch. This is releng-puppet2 in mdc1, /etc/puppet/environments/<envname>. I just do a `git clone`, `git fetch -a`, and `git update origin/<branch>`.
1. use [this command](https://wiki.mozilla.org/ReleaseEngineering/PuppetAgain/HowTo/Set_up_a_user_environment#On_the_worker_node.28s.29) to pin the machine

This allows me to a) test my patches before review, b) not have to worry if signing11 will fall through to the `signing\d+` regex, c) iterate on bustage without further review, etc etc.

Once you have a working patchset, you can send the non-pinning portion for review. Once you're ready to unpin, you can update `/etc/puppet/environments/callek` back to master, and the pinned machine(s) will unpin the next puppet run.

Callek:
> Do  we need libgsf{,-devel} installed on the signing server, or are we ok as long as the RPM was compiled with it?

the new osslsigncode now depends on libgsf, so we don't need to explicitly install it on the signing servers

> Do we need to restart the signing servers? I'm guessing no, but we should plan for it if needed.

No we do not (a try push yesterday which luckily grabbed signing11 had no issue signing msi as added confirmation)

>    coordinate with #releaseduty. You're on the hook so you should know when releases aren't running :)

Yep, I'll tell myself (and mention in channel)

>  inform #ci when you're going to land. We may break windows signing while this is updating, and we do on-push windows signing all the time on integration branches and Try.

Similarly wlll tell them

> We could roll this out a few signing servers at a time, or just deal with bustage in a smaller time window by landing.

I think landing and having all at once is easiest. Thanks

PR #290, opened 2018-11-06
Bug 1505000 - Fix workerType in mobile-beetmover-dev instances.

MihaiTabara:


MihaiTabara:
There were two culprits that enabled bug 1505000 to happen.

1. https://tools.taskcluster.net/provisioners/scriptworker-prov-v1/worker-types/mobile-beetmover-v1 lists the staging worker-Id which it shouldn't. This PR fixes it and allocates a separate workerType

2. https://tools.taskcluster.net/auth/clients/project%2Fmobile%2Fandroid-components%2Freleng%2Fscriptworker%2Fbeetmover%2Fdev had access to `queue:claim-work:scriptworker-prov-v1/mobile-beetmover-v*` which enabled it to query Taskcluster for tasks like the production one. I fixed this on TC side to take into account the new workerType from this PR.

PR #293, opened 2018-11-09
Bug 1505892 - roller snmp reboot with downtime

davehouse:


davehouse:
@zsoltfay how about we always add the delay between power off and on, if ssh fails?

zsoltfay:
@davehouse Sounds like something we can have. I think it would help eliminate 1 more step and bring us closer towards deciding if workers would need a physical check.

davehouse:
Thanks for asking for the change/fix btw, really makes more sense than manually doing it.

PR #295, opened 2018-11-12
Bug 1492245 - bump sigingscript in signingscriptworkers

MihaiTabara:


MihaiTabara:
@JohanLorenzo @escapewindow  is uploading to Pypi still required? AFAIK we don't do that for other *scripts. If yes, we need to update `9.4.1` too. I've uloaded the wheel in puppet but I don't have access to push to Pypi. I don't feel like I know enough about `signingscript` to bear that responsability, but am happy to do that if you think otherwise.

PR #303, opened 2018-11-21
Adds configuration for reference browser (prod and staging)

mitchhentges:
Minor note: the "staging" scope uses a key named with "dep" instead of "staging" because our two signing keys are `["prod", "dep"]`, not `["prod", "staging"]`


modules/signing_scriptworker/templates/passwords-mobile.json.erb
```diff
@@ -10,5 +10,11 @@
     ],
     "project:mobile:fenix:releng:signing:cert:release-signing": [
         ["https://autograph-external.prod.autograph.services.mozaws.net", "<%= scope.function_secret(["autograph_fenix_prod_username"]) %>", "<%= scope.function_secret(["autograph_fenix_prod_password"]) %>", ["autograph_fenix"], "autograph"]
+    ],
+    "project:mobile:reference-browser:releng:signing:cert:release-signing": [
+        ["https://autograph-external.prod.autograph.services.mozaws.net", "<%= scope.function_secret(["autograph_reference_browser_prod_username"]) %>", "<%= scope.function_secret(["autograph_reference_browser_prod_password"]) %>", ["autograph_reference_browser"], "autograph"]
```
MihaiTabara: This LGTM but we'll need to grab the credentials from SecOps and put them under `hiera` beforehand. @JohanLorenzo or myself can help with that.

Also, if we want the `reference-browser` APK be zipaligned, we might want to teach signingscript of its existence (e.g. https://github.com/mozilla-releng/signingscript/blob/master/signingscript/task.py#L31). If not, not need to touch `signingscript` at all as it'll automatically find it based on https://github.com/mozilla-releng/signingscript/blob/master/signingscript/sign.py#L83

modules/signing_scriptworker/templates/passwords-mobile.json.erb
```diff
@@ -10,5 +10,11 @@
...
+    ],
+    "project:mobile:reference-browser:staging:releng:signing:cert:release-signing": [
```
MihaiTabara: Hm, I'm not sure this is right. If we wanted "staging" environment testing, I think using `dep-signing` certificate scope is the way to go. AFAIK, that's how we do it for the rest of the products. See https://github.com/mozilla-releng/build-puppet/blob/master/modules/signing_scriptworker/templates/dep-passwords.json.erb. 

So in this case, I'd personally create a `dep-passwords-mobile.json.erb` that is a mirror of this file but with `dep-signing` scope instead. AFAIK, even the dep counterparts are also stored on the Autograph production instances, so we'll hit that anyway. This has the advantage that we can separate nicely between production signing and "staging" signing directly from the scope. We'd need dedicated clients for this one too from SecOps.

I'm curious what @JohanLorenzo thinks as I've been ramping up in this too in the past few weeks/months so I may be wrong.

modules/signing_scriptworker/templates/passwords-mobile.json.erb
JohanLorenzo: Mitch should now have the rights to put the creds into `hiera`. 

Good call on signingscript! We should eventually revisit the nomenclature to have a generic rule like `autograph_apk_*` to point to `sign_jar`. This way, we don't have to release a new version of signingscript every time we support a new browser.

modules/signing_scriptworker/templates/passwords-mobile.json.erb
JohanLorenzo: `project:mobile:reference-browser:releng:signing:cert:dep-signing` is the way to go :)

modules/signing_scriptworker/templates/passwords-mobile.json.erb
mitchhentges: I've added:
* `autograph_reference_browser_prod_username`: `autograph_reference_browser_prod_username`
* `autograph_reference_browser_prod_password`: `encrypted(autograph_reference_browser_prod_password)`

to `secrets.eyaml` on the puppet master :)

modules/signing_scriptworker/templates/passwords-mobile.json.erb
mitchhentges: Moved to `dep-passwords-mobile.json.erb` and renamed the properties :+1:

modules/signing_scriptworker/templates/passwords-mobile.json.erb
```diff
@@ -10,5 +10,8 @@
     ],
     "project:mobile:fenix:releng:signing:cert:release-signing": [
         ["https://autograph-external.prod.autograph.services.mozaws.net", "<%= scope.function_secret(["autograph_fenix_prod_username"]) %>", "<%= scope.function_secret(["autograph_fenix_prod_password"]) %>", ["autograph_fenix"], "autograph"]
+    ],
+    "project:mobile:reference-browser:releng:signing:cert:release-signing": [
```
MihaiTabara: This looks good, presuming we already have the credentials from SecOps. Alternatively, we can maybe feed some dummy values here so that puppet doesn't break because hiera can't find these values, until Wednesday.

modules/signing_scriptworker/templates/dep-passwords-mobile.json.erb
```diff
@@ -0,0 +1,5 @@
+{
+    "project:mobile:reference-browser:releng:signing:cert:dep-signing": [
```
MihaiTabara: This LGTM. If this lands now, it will be fine, but won't affect any existing instance as we don't currently have a dep mobil signing instance under `moco-nodes` configured with an `env_config` to use the `dep-passwords-mobile.erb` template file. We only have production instances right now, that look like [this](https://github.com/mozilla-releng/build-puppet/blob/master/manifests/moco-nodes.pp#L298).

modules/signing_scriptworker/templates/passwords-mobile.json.erb
mitchhentges: Maybe I should've posted this in a separate comment, but I [have set up some values in hiera :)](https://github.com/mozilla-releng/build-puppet/pull/303#discussion_r235832740)

modules/signing_scriptworker/templates/dep-passwords-mobile.json.erb
JohanLorenzo: I'd remove this file if we don't go with the staging environment. Otherwise, we should add `env_config`. Leaving this file alone may confuse our future selves

manifests/moco-nodes.pp
```diff
@@ -293,7 +293,18 @@
     include toplevel::server::signingscriptworker
 }
 
-# https://github.com/mozilla-mobile workers. The "e" in mobile was stripped out
+
+# Signing scriptworkers
+# In order to leave up to 100 workers, the "mobile" had to be shortened down to "m"
```
JohanLorenzo: I love this comment! Thank you for giving people context :)

PR #304, opened 2018-11-22
  Update balrogscript: Get rid of tools in balrog.

tomprince:


tomprince:
The package is deployed, so if this is good as is, can you land it, so it catches the beta build?

MihaiTabara:
Landed!

PR #312, opened 2018-11-28
Adds mobile-dep pushapk env config

mitchhentges:
I didn't updated `pushapk_scriptworker/<snip>/settings.pp` as part of #303, but here it is now!

### DIFF
--- a/modules/pushapk_scriptworker/manifests/settings.pp
+++ b/modules/pushapk_scriptworker/manifests/settings.pp
@@ -41,6 +41,7 @@ class pushapk_scriptworker::settings {
         worker_group             => 'mobile-pushapk-dep-v1',
         worker_type              => 'mobile-pushapk-dep-v1',
         verbose_logging          => true,
+        # TODO: simplify client_id to not include project ("focus")
         taskcluster_client_id    => 'project/mobile/focus/releng/scriptworker/pushapk/dep',
         taskcluster_access_token => secret('pushapk_scriptworker_taskcluster_access_token_mobile_dep'),
         scope_prefix             => 'project:mobile:focus:releng:googleplay:product:',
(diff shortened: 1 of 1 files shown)


modules/pushapk_scriptworker/manifests/settings.pp
```diff
@@ -37,6 +37,19 @@
         verify_chain_of_trust    => true,
         verify_cot_signature     => true,
       },
+      'mobile-dep' => {
+        worker_group             => 'mobile-pushapk-dep-v1',
+        worker_type              => 'mobile-pushapk-dep-v1',
+        verbose_logging          => true,
+        taskcluster_client_id    => 'project/mobile/focus/releng/scriptworker/pushapk/dep',
```
MihaiTabara: I wonder if we shouldn't have a different TC client here. Seems like `signing_scriptworker` has a similar definition of reusing the `focus` counterpart but I'm not sure why that is.

modules/pushapk_scriptworker/manifests/settings.pp
```diff
@@ -37,6 +37,19 @@
...
+        taskcluster_access_token => secret('pushapk_scriptworker_taskcluster_access_token_mobile_dep'),
+        scope_prefix             => 'project:mobile:focus:releng:googleplay:product:',
```
MihaiTabara: I'm tempted to say s/`focus`/`reference-browser` but let's see what the conclusions of the Slack thread on this is.

mitchhentges: Until https://github.com/mozilla-releng/pushapkscript/pull/54 gets merged, I can't add `project:mobile:reference-browser:...` to the `mobile-prod` env config.
My other option would be to implement this _and_ change `reference-browser` to have different pushapk scopes for dep and prod:
* `project:mobile:reference-browser:...` for dep
* `project:mobile:focus:...` for prod

But imho this looks more typo-y and confusing to future readers than using `focus` across the board (in the same way that we masquerade as `android-components`).
Either way, https://github.com/mozilla-releng/pushapkscript/pull/54 will allow us to do this without compromise

JohanLorenzo: Correct. Please keep it this way, for now :)

modules/pushapk_scriptworker/manifests/settings.pp
JohanLorenzo: Yeah, although `project:mobile:{product}` makes sense, client IDs like this one could be simplified to `project/mobile/releng/[...]`.

Let's keep it this way for now and add a comment to warn future readers.

PR #316, opened 2018-11-29
Bug 1511145: don't handle Firefox in releaserunner3

rail:
Using fake values in case we don't hablde empty arrays properly


rail:
Just prepping for now

PR #323, opened 2018-12-14
Updates Pushapk dep g_p_a_config_content to always reference g_p_config

mitchhentges:
`$google_play_accounts_config_content` should always refer to content from `$google_play_config`, even for dep.
The issue this resolves is that `dep`'s `certificate` should point to a real file, not "dummy", and the `dep` `$google_play_config` properly specifies a real file

### DIFF
--- a/modules/pushapk_scriptworker/manifests/settings.pp
+++ b/modules/pushapk_scriptworker/manifests/settings.pp
@@ -100,8 +100,8 @@ class pushapk_scriptworker::settings {
             }
             $google_play_accounts_config_content = {
                 'dep' => {
-                  'service_account' => 'dummy',
-                  'certificate' => 'dummy',
+                  'service_account' => $google_play_config['dep']['service_account'],
+                  'certificate' => $google_play_config['dep']['certificate_target_location'],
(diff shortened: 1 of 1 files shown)


modules/pushapk_scriptworker/manifests/settings.pp
```diff
@@ -100,8 +100,8 @@
             }
             $google_play_accounts_config_content = {
                 'dep' => {
-                  'service_account' => 'dummy',
-                  'certificate' => 'dummy',
+                  'service_account' => $google_play_config['dep']['service_account'],
```
JohanLorenzo: Indeed, the value is already defined to be "dummy" here https://github.com/mozilla-releng/build-puppet/pull/323/files#diff-aa7dacd4373035e75425f8728aca336dL96
I'm sorry, I should have double checked before asking https://github.com/mozilla-releng/build-puppet/pull/317#discussion_r237870273

mitchhentges: No worries, it was a team-effort "oops" :)

PR #334, opened 2018-12-19
Bumps pushapkscript and mozapkpublisher versions

mitchhentges:


modules/pushapk_scriptworker/files/requirements.txt
```diff
@@ -71,4 +71,4 @@ virtualenv==16.1.0
 voluptuous==0.11.5
 wcwidth==0.1.7
 yarl==1.3.0
-pushapkscript==0.9.0  # puppet: nodownload
+pushapkscript==0.10.0  # puppet: nodownload
```
mitchhentges: Unsure why just `pushapkscript` is `puppet: nodownload`'d :thinking:

Callek: Because `mozapkpublished` is on pypi (https://pypi.org/project/mozapkpublisher/#history) and pushapkscript isn't, so we don't have puppet download from pypi (storing local copies) of the latter.

JohanLorenzo: Hmmm, we could probably `puppet: nodownload`'d `mozapkpublisher` too. I don't remember what's the value added by this check. Would you have some pointer @escapewindow ?

mitchhentges: FWIW, @Callek, both packages are on Pypi:
* [mozapkpublisher](https://pypi.org/project/mozapkpublisher/)
* [pushapkscript](https://pypi.org/project/pushapkscript/)

mitchhentges:
I'll merge this guy once he's green, and we can come back to `nodownload` in a follow-up :)

modules/pushapk_scriptworker/files/requirements.txt
escapewindow: `# nodownload` is to avoid trying to download packages that aren't on pypi. If both of them have their latest releases on pypi and will continue to do so, we can remove the `# nodownload`.

PR #345, opened 2018-12-24
[suggestion] Adds mhentges to shortlist

mitchhentges:
This isn't _currently_ affecting me.

However, if I need to debug issues/test configuration with the signing workers, I can't ssh into them to apply them to my custom environment.
If I was on this shortlist, this would be possible.

**Alternatives:**
* Spin up a new machine (e.g.: `dep-m-signing-linux-$n`) whenever I need to apply my custom environment to a machine I don't have SSH access to
* Bug a releng-er with access to pin my environment to the machine on behalf of me, then I'll set it back after
* Get PR into `build-puppet` to pin my environment to machine (still blocked on another relenger), wait for changes to propagate, perform my debugging/testing, PR into `build-puppet` to remove pin after

### DIFF
--- a/manifests/moco-config.pp
+++ b/manifests/moco-config.pp
@@ -207,6 +207,7 @@ class config inherits config::base {
         'catlee',
         'jlorenzo',
         'jwood',
+        'mhentges',
         'mtabara',
         'nthomas',
         'raliiev',


tomprince:
One thing to note is that this also provides access to the signing servers, as well as the signing workers.

JohanLorenzo:
I agree on providing Mitch access to the signing scriptworker instances.  

Although, I'm not a big fan of providing access to the signing servers as we're moving off them. I don't think @mitchhentges needs this access. Thank you @tomprince for calling this out. @escapewindow do you think we should revisit the puppet logic to have 2 shortlists:

* one for for signing script_workers
* the other for signing servers?

escapewindow:
Hm. How about mobile shortlist vs firefox shortlist? We could add the former and have it only apply to mobile signing scriptworkers. If we’re only using autograph for mobile signing, we don’t expose signing server access.

JohanLorenzo:
Implementing the mobile shortlist sounds great to me!

mitchhentges:
I'll look into the separate shortlist, thanks!

PR #348, opened 2018-12-31
Don't explicitly set "schema_file" for scriptworker scripts

mitchhentges:
I'm nervous about the large `shipitscript` version bump - this ticket is at some risk to cause breakage

### DIFF
--- a/modules/pushsnap_scriptworker/manifests/settings.pp
+++ b/modules/pushsnap_scriptworker/manifests/settings.pp
@@ -8,7 +8,6 @@ class pushsnap_scriptworker::settings {
     include python3::settings
 
     $root                       = $config::scriptworker_root
-    $python3_virtualenv_version = $python3::settings::python3_virtualenv_version
 
     $work_dir                 = "${root}/work"
     $task_script              = "${root}/bin/pushsnapscript"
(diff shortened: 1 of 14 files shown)


mitchhentges:
There were some 5 scripts that needed updates, but they just got published in the last hour.
Thanks for the review! I'm going to wait for an additional once-over by Mihai or Johan before I merge

modules/pushsnap_scriptworker/manifests/settings.pp
```diff
@@ -72,7 +71,6 @@
 
     $script_config_content      = {
         work_dir                 => $work_dir,
-        schema_file              => "${root}/lib/python${python3_virtualenv_version}/site-packages/pushsnapscript/data/push_snap_task_schema.json",
```
JohanLorenzo: Blocker: I think you know it already since you opened https://github.com/mozilla-releng/pushsnapscript/pull/16 => we need to release a new version of pushsnapscript

mitchhentges: Fortunately, `pushsnapscript` already had a default `schema_file` set. [This PR](https://github.com/mozilla-releng/pushsnapscript/pull/16) removed `schema_file` from the example config, but didn't change any underlying logic :smile:

PR #351, opened 2019-01-02
only look in /etc/puppet/production for requirements files

escapewindow:


escapewindow:
[before and after filelists](https://gist.github.com/escapewindow/336b6fd939a9feca1db9f124e6bb68d9)

tomprince:
Does this mean that we'll need to upload files manually, if we want to test new dependencies out?

escapewindow:
That's correct. No automatic downloads for user envs.

tomprince:
Rather than not downloading for user env's, could we instead email just that user, if there is a failure in their env?

escapewindow:
That's possible, but will take a larger patch.

PR #353, opened 2019-01-03
Adds more-open mobile_shortlist for mobile signing workers

mitchhentges:
`mobile_shortlist` is equivalent to `shortlist`, plus `mhentges` added

### DIFF
--- a/manifests/moco-config.pp
+++ b/manifests/moco-config.pp
@@ -213,9 +213,29 @@ class config inherits config::base {
         'sfraser',
     ]
 
+    $mobile_shortlist = [
+        # relops
+        'klibby',
+        'jwatkins',
+        'dhouse',
+
+        # releng
+        'asasaki',
+        'bhearsum',
+        'catlee',
+        'jlorenzo',
+        'jwood',
+        'mhentges',
+        'mtabara',
+        'nthomas',
+        'raliiev',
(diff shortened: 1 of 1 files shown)


manifests/moco-config.pp
```diff
@@ -213,9 +213,29 @@
...
+        'nthomas',
+        'raliiev',
+        'sfraser',
+    ]
+
     $admin_users                     = $::fqdn ? {
         /^rejh\d\.srv\.releng\.(mdc1|mdc2)\.mozilla.com/  => $jumphost_admin_users,
         # signing machines have a very limited access list
+        /^(dep-m|mobil)-signing.*/                             => $mobile_shortlist,
         /^.*sign.*/                                            => $shortlist,
```
escapewindow: We may have to change this to `/^(?!dep-m|mobil-).*sign.*/` per https://stackoverflow.com/a/2078953 . (I don't know if these regexes are sorted, so we might hit the `/^.*sign.*/` regex before the new mobile one.)

mitchhentges: [Puppet evaluates these in order](https://puppet.com/docs/puppet/4.9/lang_conditional.html#behavior-3), but I'm :100: on being more explicit

mitchhentges:
Thanks!

PR #356, opened 2019-01-05
Adds autograph focus dep creds

mitchhentges:
[bug 1517851](https://bugzilla.mozilla.org/show_bug.cgi?id=1517851)


modules/signing_scriptworker/templates/dep-passwords-mobile.json.erb
```diff
@@ -1,4 +1,7 @@
 {
+    "project:mobile:focus:releng:signing:cert:dep-signing": [
+        ["https://autograph-external.prod.autograph.services.mozaws.net", "<%= scope.function_secret(["autograph_focus_dep_username"]) %>", "<%= scope.function_secret(["autograph_focus_dep_password" %>", ["autograph_focus"], "autograph"]
```
MihaiTabara: I think at some point we'll have to migrate these scopes to `autograph_apk_focus`. But since production looks the same in [here](https://github.com/mozilla-releng/build-puppet/blob/master/modules/signing_scriptworker/templates/passwords-mobile.json.erb#L3), this LGTM for now. We can address it later when we'll address `fenix` too.

PR #398, opened 2019-02-19
Bumps scriptworker version to 19.0.0

mitchhentges:
Scriptworker now reports `worker-shutdown` if the task is terminated.
[`19.0.0` changelog](https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#1900---2019-02-13)

### DIFF
--- a/modules/addon_scriptworker/files/requirements.txt
+++ b/modules/addon_scriptworker/files/requirements.txt
@@ -26,7 +26,7 @@ python-gnupg==0.4.4
 python-jose==3.0.1
 requests==2.21.0
 rsa==4.0
-scriptworker==18.1.0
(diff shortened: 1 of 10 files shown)


modules/addon_scriptworker/files/requirements.txt
```diff
@@ -26,7 +26,7 @@ python-gnupg==0.4.4
 python-jose==3.0.1
 requests==2.21.0
 rsa==4.0
-scriptworker==18.1.0
+scriptworker==19.0.018.1.0
```
escapewindow: I think you mean `19.0.0`, not `19.0.018.1.0` ?

PR #405, opened 2019-02-27
Bug 1530952 - Upgrade to taskcluster-proxy 5.1.0 and generic-worker 13.0.2 on Linux and OSX workers

dragoscrisan:
Upgrade generic-worker to v13.0.2 and taskcluster-proxy to v5.1.0 on Linux and OSX workers


modules/generic_worker/templates/generic-worker.config.erb
```diff
@@ -20,7 +20,7 @@
   "privateIP": "",
   "provisionerId": "releng-hardware",
   "publicIP": "<%= @ipaddress %>",
-  "refreshURLsPrematurelySecs": 310,
+  "rootURL": "https://taskcluster.net",
```
petemoore: nit: it would be nicer to preserve alphabetical ordering, and put `rootURL` after `requiredDiskSpaceMegabytes`.

PR #408, opened 2019-02-27
Bumps beetmover max zip file size to 100mb

mitchhentges:
This is for [this ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1522581). An earlier fix for the issue bumped the default maximum file size, but we missed the explicit override in puppet.

Since we don't _want_ a different maximum size, perhaps we can just lean on the default instead?
(If we'd rather be explicit in puppet, we should remove the default from `beetmover` and make it "required", imho)

### DIFF
--- a/modules/beetmover_scriptworker/templates/script_config.json.erb
+++ b/modules/beetmover_scriptworker/templates/script_config.json.erb
@@ -10,7 +10,7 @@
 
     "verbose": <%= scope.lookupvar("beetmover_scriptworker::settings::verbose_logging") %>,
 
-    "zip_max_file_size_in_mb": 100,
+    "zip_max_file_size_in_mb": 300,
 
     "bucket_config": {
 <% if @env_config["nightly_buckets"] -%>


escapewindow:
The model I like to use in puppet is,

1. bump the config in puppet
2. bump the default in the \*script or scriptworker
3. roll that default bump out in the next release, whenever that is

The downside here is clear: we have to change things in two places. Two different PRs. However, the upside is:

1. bumping it in puppet is faster than bumping it in the \*script, releasing the script, uploading the \*script to puppet, bumping the version in puppet, potentially dealing with any other ridealongs in the new release that we aren't ready for yet, etc. etc.
2. having the default in source allows us to test and use the script without puppet.

I think the pluses are greater than the minuses. I'd lean towards bumping the default here to 300 rather than removing.

mitchhentges:
Ahh, good point! Modified accordingly :)

-----

FWIW, we could have the same benefits (with a minor decrease in runtime logic) if we kept the default value in [`config_example.json`](https://github.com/mozilla-releng/beetmoverscript/blob/master/config_example.json#L59), but _required_ a value at runtime.

This would require `config.json` to _always_ be explicit, but we're the sole consumer, and we want it to be explicit already, so that might not be a downside.

We wouldn't gain much from this, but it's mildly interesting :stuck_out_tongue:

escapewindow:
Sure. If we have any integration or other tests, we also have to provide an explicit value there as well, and we may have to bump that test default along with the `config_example.json` default and puppet runtime value. Roughly the same in my mind, but both should work.

mitchhentges:
On the other hand, we get to remove tests that verified the old functionality. Also, IMHO, it's a lesser cost to have tests with duplicated configuration values than runtime logic.
Fun food-for-thought, but not really that impactful. Thanks for joining my hypothetical adventure :smile:

PR #409, opened 2019-02-28
Bug 1507408 - Refactoring automatic bug generation

dragoscrisan:
With the changes to the workflow that are coming very soon, the automated bug generator will need to suffer some modifications which I have covered with :dragrom in a meeting.

To outline the points discussed:
- Bug Title should be under the following format: [$(MDC)] $(HOSTNAME) Generic Worker CODE 69.
-- End result should look like: [MDC1] t-yosemite-r7-XXX Generic Worker CODE 69

- Add an Alias of $(HOSTNAME) to ensure we don't have duplicated bugs for the same machine. If that alias already exists, use the existing Problem Tracking bug .

- If possible, automatically add CiDuty's team account ( https://bugzilla.mozilla.org/user_profile?login=ciduty%40mozilla.com )to NeedInfo (everyone in the team has visibility and will receive the email from NI?)


dragoscrisan:
This bug was created using this script: https://bugzilla.mozilla.org/show_bug.cgi?id=1534923

davehouse:
This refactoring looks good and improves the readibility of the bugzilla-bug creation.

I'm okay with it despite my concern about adding a python process as a parent monitoring generic-worker; I understand the motivation to improve the code and for maintenance (python is more clear than a shell script for a complicated exit code handling and bug creation).

modules/generic_worker/templates/bug_utility.py.erb
```diff
@@ -0,0 +1,81 @@
...
+    bug_version = "unspecified"
+    bug_summary = "["+get_domain()+"] "+ get_hostname() + " Generic worker exit code 69"
+    bug_description = ("Error code 69 means: worker panic - either a worker bug, or the environment is not suitable for running "
+                    "a task, e.g. a file cannot be written to the file system, or something else did "
+                    "not work that was required in order to execute a task. See config setting "
+                    "shutdownMachineOnInternalError. "
+                    "The following log were provided: "
+                    "https://papertrailapp.com/systems/"+get_fullname()+"/events?q=program%3Ageneric-worker")
+    bug_info = bugzilla.build_createbug(
+        alias       = bug_alias,
```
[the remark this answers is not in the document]
davehouse: This is resolved by changing the alias to hostname only. Thanks

modules/generic_worker/templates/run-generic-worker.py.erb
```diff
@@ -0,0 +1,101 @@
+#!/usr/local/bin/python3
+
+import sys
+import imp
```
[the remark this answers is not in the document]
dragoscrisan: We can use the following approach: 
1. Run generic-worker from shell script (like now)
2. use python to treat generic-worker exit code and open/close bugs

I really want to keep run-generic-worker shell scrip simple, something like this:

rm -f tasks-resolved-count.txt

export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin
source /usr/local/share/generic-worker/bugzilla-utils.sh

# First run the generic-worker, passing through any arguments handed to this
# wrapper script...
/usr/local/bin/generic-worker "$@" 2>&1 | logger -t generic-worker -s
exitstatus="${PIPESTATUS[0]}"

python3 generic-worker-utils.py $exitstatus

Dave, are you OK with this approach?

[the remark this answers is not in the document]
davehouse: +1
I think this is more robust because it will still work if we have python issues (or other problems in the exit handling like with bugzilla), and we won't have the python interpreter as a parent for the worker (may not be a security or performance issue, but it feels like it to me ;).

To me, this reads like "always run generic-worker, and do something based on the exit code". Does that match the purpose?

You could also remove the sourcing of the bugzilla-utils.sh, and the rm of tasks-resolved-count.txt (that is only used by the python script, right?).

modules/generic_worker/templates/generic-worker-utils.py.erb
```diff
@@ -0,0 +1,84 @@
...
+
+def main():
+    # This is a wrapper script for calling generic worker, such that when the worker
+    # exits, this script will reboot the machine. Note, in the worker config
+    # (/etc/generic-worker.config) we set numberOfTasksToRun to 1 so that the worker
+    # will exit after running a single task. This script is then responsible for
+    # rebooting the machine.    
+    quarantine_file = "<%= scope.lookupvar('users::builder::home') %>/quarantined"
+
+    # Chex the exitcode for generic-worker process
```
petemoore: nit: Checks

modules/generic_worker/templates/generic-worker-utils.py.erb
```diff
@@ -0,0 +1,84 @@
...
+    exit_code = int(sys.argv[1])
+    if exit_code in get_exitcodes_dictionary():
+        # If generic-worker considers that the host environment is in an invalid/corrupted
+        # state (for example, it cannot write to disk), it will return an exit code other that 0, 67 or 68.
```
petemoore: nit: other _than_ (and line wrapping)

modules/generic_worker/templates/generic-worker-utils.py.erb
```diff
@@ -0,0 +1,84 @@
...
+        #
+        # More information about generic-worker exit codes can be seen here:
+        # https://docs.taskcluster.net/docs/reference/workers/generic-worker#set-up-your-env
```
petemoore: nice

modules/generic_worker/templates/generic-worker-utils.py.erb
Akhliskun: Resolved in commit 32f711e

modules/generic_worker/templates/generic-worker-utils.py.erb
Akhliskun: Resolved in commit 32f711e

modules/generic_worker/templates/generic-worker-utils.py.erb
Akhliskun: Resolved in commit 32f711e

modules/generic_worker/manifests/init.pp
```diff
@@ -54,6 +54,10 @@
             # The reboot command in OSX not have --force option
             $reboot_command = '/usr/bin/sudo /sbin/reboot'
 
+            # Since python 3.7.1 require openssl 1.0.2 or greater, and on OSX workers we have openssl 0.98
+            # and python3 is pinned I'll use python3.6 instead python3
+            $python3 = '/usr/local/bin/python3.6'
```
davehouse: :crying_cat_face:

davehouse: Long-term, it may be better to move the quarantine-worker to be a statically compiled binary or to run with an interpreter that is available everywhere (or maybe python code that is backwards-compatible).

davehouse:
I've been thinking about this more related to the mass-quarantines last week (re: bug https://bugzilla.mozilla.org/show_bug.cgi?id=1537886 asking if we can handle that better). And I'd like for it to require less interaction (emails, bugs, un-quarantine) and be easier to understand and manage. The code looks good, and I like that it is automating the handling of problems. And I think we can improve this further.

Could we meet next month and plan the next iteration of this? @dividehex and @petemoore I'd like your input for v2 also.

dividehex:
> I've been thinking about this more related to the mass-quarantines last week (re: bug https://bugzilla.mozilla.org/show_bug.cgi?id=1537886 asking if we can handle that better). And I'd like for it to require less interaction (emails, bugs, un-quarantine) and be easier to understand and manage. The code looks good, and I like that it is automating the handling of problems. And I think we can improve this further.
> 
> Could we meet next month and plan the next iteration of this? @dividehex and @petemoore I'd like your input for v2 also.

I've been thinking about this also and I think we might be introducing too much complexity into the issue especially now that it looks like we are switching the control script to python and there are lots of dependencies being introduced.  I also think the need to build packages here ultimately adds unnecessary complexity and probably isn't sustainable in the long run.  I think we should definitely meet to talk about the 30,000 foot view of what we are trying to achieve here.

petemoore:
@dividehex should we close this now?

grenade:
closed at pete's request

PR #425, opened 2019-03-11
 Bug 1129103 - Verify that OpenSSH configuration complies with guidelines

dragoscrisan:
Verify that OpenSSH configuration complies with guidelines

### DIFF
--- a/modules/ssh/templates/sshd_config_duo.erb
+++ b/modules/ssh/templates/sshd_config_duo.erb
@@ -17,9 +17,9 @@ PermitRootLogin no
 PasswordAuthentication no
 
 # Required for Duo MFA
-AuthenticationMethods publickey,keyboard-interactive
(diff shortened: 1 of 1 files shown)


dragoscrisan:
For ssh config duo, these are configurations reccommanded by infosec:
# IMPORTANT: you will have to ensure OpenSSH cannot authenticate with passwords with PAM in /etc/pam.d/sshd
# "PasswordAuthentication no" is not sufficient!
PubkeyAuthentication yes
PasswordAuthentication no
AuthenticationMethods publickey,keyboard-interactive:pam
KbdInteractiveAuthentication yes
UsePAM yes
# Ensure /bin/login is not used so that it cannot bypass PAM settings for sshd.
# Note, this option is no longer needed as of OpenSSH 7.4 as support for UseLogin has been removed
UseLogin no

See: https://infosec.mozilla.org/guidelines/openssh

PR #439, opened 2019-03-28
scriptworker 23.0.0 - remove gpg support

escapewindow:


escapewindow:
I managed to get this onto signing-linux-dev1 via my puppet env. It removed the cron entry, claimed a task, and went green.

modules/scriptworker/manifests/chain_of_trust.pp
```diff
@@ -32,47 +16,9 @@
...
-          command   => "${basedir}/bin/rebuild_gpg_homedirs ${basedir}/scriptworker.yaml",
-          subscribe => File[$git_pubkey_dir],
-          user      => $username;
-      # Create checksum file of git pubkeys
-      "${basedir}/.git-pubkey-dir-checksum":
-          require => File[$git_pubkey_dir],
-          path    => '/usr/local/bin/:/bin:/usr/sbin:/usr/bin',
-          user    => $username,
-          command => "find ${git_pubkey_dir} -type f | xargs md5sum | sort > ${basedir}/.git-pubkey-dir-checksum";
-  }
```
tomprince: I think all the files referenced here (both directly, and via exec) probably want to be marked as absent?

escapewindow: I thought about that. I leaned towards keeping them, and getting rid of the last traces by migrating to docker/GCP. Because they're no longer valid keys according to scriptworker, I'm less concerned here. I can change the patch to specify `absent`, but I'll have to keep the configs in place to do so.

tomprince: It *looks* like all the files are fixed relative to the basedir, so you could hard-code the things to remove? I don't think anything actually uses the flexibility to specify them individually.

tomprince:
There are also the the two entries in moco-config.pp to be removed.

modules/scriptworker/manifests/chain_of_trust.pp
escapewindow: Smart. https://grokbase.com/t/gg/puppet-users/1446kqrt5g/how-to-remove-a-directory#201404083akctk6dr7pmqf5cndxfste3ky worked!

escapewindow:
I was able to test file cleanup on signing-linux-dev1; worked! Added hiera cleanup to the bug checklist, and I addressed the moco-config.pp entries. It's too late to roll out today, so if this passes review, I'll probably aim for Thursday after we're done with releases. Thanks!

PR #446, opened 2019-04-08
Telegraf on hw workers

davehouse:
Adding the telegraf daemon to the hardware workers.
1. adds telegraf config into toplevel/worker
2. updates telegraf module for MacOS install and service control
3. adds a separate user for telegraf to run under
4. includes the build script for the MacOS dmg

### DIFF
--- a/modules/users/manifests/telegraf.pp
+++ b/modules/users/manifests/telegraf.pp
@@ -1,0 +1,23 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class users::telegraf {
+    $username = 'telegraf'
+    $home = $::operatingsystem ? {
+        Darwin  => "/Users/${username}",
+        default => "/home/${username}"
+    }
+    $group = $::operatingsystem ? {
+        Darwin  => 'staff',
+        Ubuntu  => undef,
+        default => $username
+    }
+
+    user {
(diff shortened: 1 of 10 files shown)


modules/users/manifests/telegraf.pp
```diff
@@ -0,0 +1,23 @@
...
+    $group = $::operatingsystem ? {
+        Darwin  => 'staff',
+        Ubuntu  => undef,
+        default => $username
+    }
+
+    user {
+        'telegraf':
+            # password   => '*',  # none allowed
+            groups  => $group,
```
dividehex: The 'gid' attrib should probably be used instead of 'groups' since it is the primary group for the user.

modules/users/manifests/telegraf.pp
```diff
@@ -0,0 +1,23 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+class users::telegraf {
+    $username = 'telegraf'
+    $home = $::operatingsystem ? {
```
dividehex: Why set $username and $home if they are unused?  Just remove them.

modules/users/manifests/telegraf.pp
davehouse: thx. I switched to using gid.

davehouse:
@dividehex r? I had updated this to use the gid and home and tested it on a handful of linux and macos workers.

PR #455, opened 2019-04-18
Add appservices support. Bump scriptworker 23.0.2

MihaiTabara:


MihaiTabara:
@JohanLorenzo Is the TC ci failing because `23.0.2` is missing from Pypi? If so, can you please add it there? Or grant me permissions to push, either wfm 👍

PR #461, opened 2019-04-24
Bump balrogclient version to pick up fix

mozbhearsum:


mozbhearsum:
D'oh, I didn't realize `balrogclient` had to be on pypi. I did that, let's fix CI....

PR #470, opened 2019-05-07
Support for new signingscript with gpg, widevine and mar keyid support

catlee:


modules/signing_scriptworker/files/requirements.txt
```diff
@@ -47,4 +47,5 @@ uritemplate==3.0.0
 urllib3==1.24.2  # pyup: ignore
 virtualenv==16.5.0
 wheel==0.33.1
+widevine==0.1.0  # puppet: nodownload
```
escapewindow: Nit, let's start using [version 1.0.0+ for things in production](https://github.com/mozilla-releng/releng-rfcs/blob/master/rfcs/0014-use-production-versions-for-scripts.md). We can follow up with this, especially if we're not live with production widevine signing yet.

PR #472, opened 2019-05-08
Removes Fenix from jarsigner aliases

mitchhentges:
This is needed for the Fenix Beta to be pushed to the Google Play Store

Details: Allows Fenix to supply the certificate alias as a task payload, so `pushapk_scriptworker` doesn't infer it from scopes.

Bugging you Aki because Johan is away and you're a cool guy :)

### DIFF
--- a/modules/pushapk_scriptworker/files/requirements.txt
+++ b/modules/pushapk_scriptworker/files/requirements.txt
@@ -52,7 +52,7 @@ pexpect==4.7.0
 pickleshare==0.7.5
 prompt_toolkit==2.0.9
 ptyprocess==0.6.0
-pushapkscript==1.0.2  # puppet: nodownload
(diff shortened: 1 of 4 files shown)


mitchhentges:
This must be deployed at the same time as [this PR](https://github.com/mozilla-mobile/fenix/pull/2349), otherwise nightlies will break

escapewindow:
To clarify, the pushapk bump is what enables Fenix Beta, and removing the fenix configs is just cleanup?

mitchhentges:
The pushapk bump is associated with a change in how the certificate alias is calculated:
1. Read from `jarsigner_aliases`, using the android product as the key
2. Read from the task payload `certificate_alias`
3. If both are provided, fail the builds due to ambiguity

Removing the fenix configs in this PR is mandatory, otherwise the pushapk build will fail due to having an ambiguous alias

PR #475, opened 2019-05-11
Resolve fenix nightlies failing

mitchhentges:
This does three things:
* Updates `pushapkscript` so that using `certificate_alias` doesn't throw a schema-validation error
* Adds `has_nightly_track` to configs that were missing it, since it's required
* Adds a base `fenix` product config because it's inferred from scope, and we don't have separate `project:mobile:...:fenix-$channel` scopes

### DIFF
--- a/modules/pushapk_scriptworker/manifests/settings.pp
+++ b/modules/pushapk_scriptworker/manifests/settings.pp
@@ -231,6 +231,11 @@ class pushapk_scriptworker::settings {
         }
         'mobile-prod': {
             $google_play_config = {
+                'fenix' => { # TODO replace with fenix-nightly
+                    service_account             => $_google_play_accounts['fenix-nightly']['service_account'],
+                    certificate                 => $_google_play_accounts['fenix-nightly']['certificate'],
+                    certificate_target_location => "${root}/fenix-nightly.p12",
+                },
                 'fenix-nightly' => {
                     service_account             => $_google_play_accounts['fenix-nightly']['service_account'],
                     certificate                 => $_google_play_accounts['fenix-nightly']['certificate'],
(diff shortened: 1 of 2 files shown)


modules/pushapk_scriptworker/manifests/settings.pp
```diff
@@ -253,8 +258,21 @@
                 },
             }
             $product_config = {
+                'fenix' => { # TODO replace with fenix-nightly
```
MihaiTabara: I'm somewhat confused by the `has_nightly_track` flag. While for `fenix` project it makes sense somehow, for `fenix-beta` and `fenix-nightly` it feels like it's adding some confusion. The former makes no sense to have a nighly channel since it's another type of shipping channel as well, while the latter is itself it 🤔

mitchhentges: I agree that it's bizarre - I think we're in an "in-between" state right now where we _need_ different config for each channel (e.g.: `expected_package_names`), so we need separate configs (`fenix`, `fenix-nightly`, `fenix-beta`) yet our structure right now is built around single-config-per-product.

One Fenix is stable again, I'm considering modifying this like:
```
fenix => {
    'update_google_play_strings' => false,
    'digest_algorithm' => 'SHA-256',
    'skip_...' => true,
    'channels' => [{
        'track' => 'nightly',
        'expected_package_name' => 'org.mozilla.fenix.nightly',
    }, {
        'track' => 'beta',
        'expected_package_name' => 'org.mozilla.fenix.beta',
    }]
}
```

MihaiTabara: This looks much better, indeed! Sounds good, we can maybe file a github issue or a bug or alike to track this once the fenix is in a more stable state 👍

mitchhentges:
ez: https://taskcluster-web.netlify.com/tasks/HYp0h8dNSluEBxzkD-xZJA

modules/pushapk_scriptworker/manifests/settings.pp
MihaiTabara: Thanks :)

PR #478, opened 2019-05-13
Disable shipitworkers

rail:
They have been working fine in GCP/cloudops for a few weeks now.


rail:
> We should remove modules/shipit_scriptworker as well. I'm ok if that's in a followup.

I haven't moved the Thunderbird scriptworkers yet.

escapewindow:
Aha! Makes sense.

PR #499, opened 2019-06-03
Fenix production signing

mitchhentges:
This will:
1. Enable us to release to the third Fenix track (nightly) in the future
2. Be more clear that we're using the autograph "nightly" signing keys for production, now


modules/pushapk_scriptworker/manifests/init.pp
```diff
@@ -112,6 +112,8 @@
                     content => $google_play_accounts['fenix-nightly']['certificate'];
                 "${root}/fenix_beta.p12":
                     content => $google_play_accounts['fenix-beta']['certificate'];
+                "${root}/fenix_production.p12":
```
MihaiTabara: Should we name this `fenix_release` to be consistent? I see `focus` has something similar `focus_release`...

modules/pushapk_scriptworker/manifests/settings.pp
```diff
@@ -219,6 +219,13 @@
                             'google_credentials_file' => "${root}/fenix_beta.p12",
                             'certificate_alias' => 'fenix-beta',
                         },
+                        'production' => {
```
MihaiTabara: Same question as above, any reason why we're not opting for `fenix-release` here?

modules/pushapk_scriptworker/manifests/init.pp
mitchhentges: Good question: `focus` uses a different deployment scheme - though it has multiple channels (nightly/production), they're all represented under the same Google Play Store entry. So, they all have the same package name, and are signed with the same key.

So, for terminology: if an app is made available for real customers to use (e.g.: the channels nightly/beta/production), then the app is "released". This makes builds for these channels "release builds". Since Focus uses the same signing/push credentials for all of its channels, then we can name the credentials accordingly: `focus_release`

Fenix is different, because it has a different app for every channel. So, we need separate credentials for our release channels. I have named each credential to match the channel name (nightly/beta/production)

modules/pushapk_scriptworker/manifests/settings.pp
mitchhentges: You mentioned in slack if "production" made sense as the name of the "main" channel.
I've found that there's a general convention to name the product's "main" channel as "production". Usually your other releases will have names like "nightly", "staging", "beta", "qa", and more, but the primary environment/channel tends to be "production".

If your product only has one release channel (there's no beta/staging/etc), then sometimes it will just be called "release".

This lines up with `gradle` configuration, which expects only a single release channel by default, so calls it `release` (e.g.: `assembleRelease`). Now that we've configured `gradle` for each of our three channels (nightly/beta/production), the term "release" is too general to refer to our "main" release.

I hope this makes sense, let me know if you want to chat :)

PR #504, opened 2019-06-06
Bug 1554769: Update KEY file with the new public key

catlee:


catlee:
Used https://gist.github.com/catlee/8eb617be64a297f1ab1a157b7543e9d8 to test.

PR #520, opened 2019-06-18
Updates balrogscript and runs in py3 environment

mitchhentges:
### Merge checklist

* [x] Run staging release with `balrogscript` running in Py3

### DIFF
--- a/modules/balrog_scriptworker/manifests/init.pp
+++ b/modules/balrog_scriptworker/manifests/init.pp
@@ -7,7 +7,6 @@ class balrog_scriptworker {
     include dirs::builds
     include packages::mozilla::python3
     include packages::mozilla::python27
-    include packages::mozilla::py27_mercurial
     include users::builder
     include tweaks::swap_on_instance_storage
     include packages::gcc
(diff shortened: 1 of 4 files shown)


modules/balrog_scriptworker/manifests/init.pp
```diff
@@ -35,17 +35,7 @@
             user            => $users::builder::username,
             group           => $users::builder::group,
             mode            => '0700',
-            packages        => file("balrog_scriptworker/requirements-3.txt");
```
MihaiTabara: I suppose we can remove lines 9 and 10 as well since we no longer need python2?

modules/balrog_scriptworker/manifests/init.pp
```diff
@@ -35,17 +34,20 @@
...
     python27::virtualenv {
         "${balrog_scriptworker::settings::root}/py27venv":
             python          => $packages::mozilla::python27::python,
             rebuild_trigger => Class['packages::mozilla::python27'],
             require         => Class['packages::mozilla::python27'],
             user            => $users::builder::username,
             group           => $users::builder::group,
-            packages        => file("balrog_scriptworker/requirements-27.txt");
+            ensure          => 'absent',
+            packages        => file("balrog_scriptworker/requirements.txt");
```
JohanLorenzo: Nit: I just noticed this line may cause a conflict once deployed

mitchhentges: Ah, and leave `requirements-27` populated, but note that it's unused? :+1:

JohanLorenzo: Hmmm, good call. It's probably not worth it. Let's keep it the way it is now.

PR #534, opened 2019-07-01
Bug 1548545 - add application-services signing workers.

MihaiTabara:
Not yet to be merged.


modules/signing_scriptworker/templates/dep-passwords-appsv.json.erb
```diff
@@ -0,0 +1,5 @@
+{
+    "project:mozilla:application-services:releng:signing:cert:dep-signing": [
+        ["https://autograph-external.prod.autograph.services.mozaws.net", "signingscript_focus_dep", "<%= scope.function_secret(["autograph_focus_dep_password"]) %>", ["autograph_focus"], "autograph"]
```
MihaiTabara: TODO: do we want different credentials here for Autograph?

modules/signing_scriptworker/templates/passwords-appsv.json.erb
```diff
@@ -0,0 +1,5 @@
+{
+    "project:mozilla:application-services:releng:signing:cert:release-signing": [
```
MihaiTabara: TODO: do we want different credentials here for Autograph?

MihaiTabara:
CC: @mitchhentges

modules/signing_scriptworker/manifests/settings.pp
```diff
@@ -123,5 +123,35 @@
             datadog_api_key          => secret('scriptworker_datadog_api_key'),
             gpg_keyfile              => 'KEY_dep',
         },
+        'appsv-dep' => {
+            worker_type              => 'appsv-signing-dep-v1',
+            worker_group             => 'appsv-signing-dep-v1',
+            taskcluster_client_id    => 'project/mozilla/application-services/releng/scriptworker/signing/dep',
+            taskcluster_access_token => secret('application_services_signing_dep_scriptworker_taskcluster_access_token'),
+            passwords_template       => 'dep-passwords-appsv.json.erb',
+            scope_prefixes           => ['project:mobile:focus:releng:signing:', 'project:mobile:android-components:releng:signing:', 'project:mobile:fenix:releng:signing:', 'project:mobile:reference-browser:releng:signing:', 'project:mozilla:application-services:releng:signing'],
```
mitchhentges: We only need `project:mozilla:application-services:releng:signing` here, right?

modules/signing_scriptworker/manifests/settings.pp
```diff
@@ -123,5 +123,35 @@
...
+            taskcluster_access_token => secret('application_services_signing_scriptworker_taskcluster_access_token'),
+            passwords_template       => 'passwords-appsv.json.erb',
+            scope_prefixes           => ['project:mobile:focus:releng:signing:', 'project:mobile:android-components:releng:signing:', 'project:mobile:fenix:releng:signing:', 'project:mobile:reference-browser:releng:signing:', 'project:mozilla:application-services:releng:signing'],
+            sign_chain_of_trust      => true,
+            verify_chain_of_trust    => true,
+            verify_cot_signature     => true,
+            cot_product              => 'application-services',
+            github_oauth_token       => secret('scriptworker_github_oauth_token_production'),
+            datadog_api_key          => secret('scriptworker_datadog_api_key'),
+            gpg_keyfile              => 'KEY_dep',
```
mitchhentges: Should this be `KEY_dep`?

modules/signing_scriptworker/templates/dep-passwords-appsv.json.erb
mitchhentges: I don't think we want the _focus_ dep key (is that what we're using for `android-components`?) but I don't think that we need an `application-services`-specific key. Can we use a generic "mozilla library" dep and release signing key?

modules/signing_scriptworker/templates/passwords-appsv.json.erb
mitchhentges: Looks like this is the same key that's used by `android-components`, right? I think that re-using this key for other "mozilla library" signing should be fine.
Thinking about key management: downstream projects need to trust the signing key of the signed dependencies they use. It makes sense to have to manage different keys for different organizations, but if they need to add a new trusted signing key for each new _dependency_ (or group of dependencies) that they import, I'm not sure that that makes sense. Trusting a single "mozilla library" key makes the most sense, IMHO

modules/signing_scriptworker/manifests/settings.pp
MihaiTabara: Good catch, I addressed it in a follow-up.

modules/signing_scriptworker/manifests/settings.pp
MihaiTabara: Good catch, I addressed it in a follow-up.

modules/signing_scriptworker/templates/dep-passwords-appsv.json.erb
MihaiTabara: Good call, I pointed to the gpg default one,.

modules/signing_scriptworker/templates/passwords-appsv.json.erb
MihaiTabara: Sounds good, let's keep this as is 👍

MihaiTabara:
@mitchhentges ready for your re-review 😎

PR #539, opened 2019-07-08
Bug 1561617 - Bump {mozilla-version,treescript} to {0.3.4,1.2.1}

JohanLorenzo:


modules/mergeday/files/requirements.txt
```diff
@@ -1,2 +1,5 @@
 # python_version: 27
+attrs==19.1.0
+enum34==1.1.6
+mozilla-version==0.3.4
```
JohanLorenzo: This is now needed because of https://bugzilla.mozilla.org/attachment.cgi?id=9076779

PR #547, opened 2019-07-11
Bugfix: bump beetmoverscript in beetmoverworkers.

MihaiTabara:


modules/beetmover_scriptworker/files/requirements.txt
```diff
@@ -7,7 +7,7 @@ arrow==0.14.2
 asn1crypto==0.24.0
 async_timeout==3.0.1
 attrs==19.1.0
-beetmoverscript==8.5.1  # puppet: nodownload
+beetmoverscript==8.5.3  # puppet: nodownload
```
MihaiTabara: Note: 8.5.2 was an un-released version in puppet as Rail was testing the CloudOps deployments - https://github.com/mozilla-releng/beetmoverscript/blob/master/CHANGELOG.rst#852--2019-07-10

PR #549, opened 2019-07-12
[bug 1562061] fix darwin kill mig-agent exec

dividehex:
Fixes the kill mig darwin exec resource


dividehex:
https://bugzilla.mozilla.org/show_bug.cgi?id=1562061

PR #561, opened 2019-07-25
fix docutils

escapewindow:
https://pypi.org/project/docutils/#files

We're getting puppet mail about this. I was able to download docutils-0.15.1.post1 and that fixed a manual install, but we need to land this change to stop getting puppet errors.


escapewindow:
Thanks!

PR #576, opened 2019-08-12
Add releng bits to handle Nazgul bounceradmin both staging/prod environments

MihaiTabara:


manifests/moco-nodes.pp
```diff
@@ -441,6 +441,8 @@
 # Bouncer scriptworkers
 node /^bouncerworker-dev.*\.srv\.releng\..*\.mozilla\.com$/ {
     $aspects                  = [ 'maximum-security' ]
+    $pin_puppet_server = 'releng-puppet2.srv.releng.mdc1.mozilla.com'
```
MihaiTabara: This will be removed before asking for review.

modules/bouncer_scriptworker/files/requirements.txt
```diff
@@ -5,7 +5,7 @@ arrow==0.14.4
 asn1crypto==0.24.0
 async_timeout==3.0.1
 attrs==19.1.0
-bouncerscript==3.7.0  # puppet: nodownload
+bouncerscript==3.7.1.dev0  # puppet: nodownload
```
MihaiTabara: This will be changed to `3.7.1` once we're ready to roll-out.

modules/bouncer_scriptworker/manifests/settings.pp
```diff
@@ -21,6 +23,13 @@
         password                  => secret('ffx-bouncer-staging_password'),
     }
 
+    $bouncer_nazgul_stage_instance_config = {
+        api_root                  => 'https://bounceradmin-default.stage.mozaws.net/api',
+        timeout_in_seconds        => 60,
+        username                  => 'taskcluster',
```
MihaiTabara: The `username` is likely to change before rolling out to production.

modules/bouncer_scriptworker/manifests/settings.pp
```diff
@@ -52,13 +62,21 @@
...
                 timeout_in_seconds        => 60,
                 username                  => 'ffxbld',
                 # TODO Split credentials
                 password                  => secret('tuxedo_password'),
             },
+            'project:releng:bouncer:server:production-nazgul' => {
+                api_root                  => 'https://bounceradmin-default.prod.mozaws.net/api',
+                timeout_in_seconds        => 60,
+                username                  => 'taskcluster',
+                # TODO Split credentials
```
rail: Do you need to address this todo?

rail:
We should also replicate this in https://github.com/mozilla-releng/bouncerscript/blob/master/docker.d/worker.yml

modules/bouncer_scriptworker/manifests/settings.pp
MihaiTabara: No, I kept the copy-paste from the old one. No action for now I suppose, until we're fully migrated to the new Nazgul.

MihaiTabara:
Note to self: 
* I also need to address Thunderbird here
* potentially rename the existing usernames, they are not very consistent with what we have in the old bounceradmin

PR #584, opened 2019-08-21
bug1561040 disabling collectd requires collectd

davehouse:
bsdpyN.tier3 missing collectd complained that the service could not be turned off because it did not exist

I worked-around the problem on those hosts manually by installing collectd, but we'll need this to avoid problems in the future when collectd is not installed.

instead of doing this, we could add an exec that checks for the service and only turns it off if it exists.

### DIFF
--- a/modules/collectd/manifests/disable.pp
+++ b/modules/collectd/manifests/disable.pp
@@ -5,6 +5,9 @@
 class collectd::disable {
     include collectd::settings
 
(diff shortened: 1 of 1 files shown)


davehouse:
We could write a new ::disable that removes the package, but I think this is simpler to leave it verified-off/controlled by puppet. (and I'm viewing build-puppet as deprecated)

PR #588, opened 2019-08-29
Bug 1540276: Deploy new signingscript with authenticode signing support

catlee:


rail:
The CI check fails downloading winsign, what is ok for now.

PR #591, opened 2019-09-03
Adds signingscript configuration to sign fenix as fennec

mitchhentges:
See [this Fenix ticket for context](https://github.com/mozilla-mobile/fenix/issues/4873).

I'll update the Fenix automation to produce a signing task with a payload like:
```
# scope: "project:mobile:fenix:releng:signing:format:autograph_apk_fennec_sha1"
# scope: "project:mobile:fenix:releng:signing:cert:fennec-production-signing"
{
  "upstreamArtifacts": [
    {
      "formats": [
        "autograph_apk_fennec_sha1"
      ],
      "paths": [...],
      "taskId": "...",
      "taskType": "build"
    }
  ]
}
```

### DIFF
--- a/modules/signing_scriptworker/templates/passwords-mobile.json.erb
+++ b/modules/signing_scriptworker/templates/passwords-mobile.json.erb
@@ -11,6 +11,9 @@
     "project:mobile:fenix:releng:signing:cert:production-signing": [
         ["https://autograph-external.prod.autograph.services.mozaws.net", "signingscript_fenix_release", "<%= scope.function_secret(["autograph_fenix_production_password"]) %>", ["autograph_apk"], "autograph"]
     ],
+    "project:mobile:fenix:releng:signing:cert:fennec-production-signing": [
+        ["https://autograph-external.prod.autograph.services.mozaws.net", "signingscript_fennec_rel", "<%= scope.function_secret(["autograph_fennec_release_password"]) %>", ["autograph_apk"], "autograph"]
+    ],
     "project:mobile:reference-browser:releng:signing:cert:release-signing": [
         ["https://autograph-external.prod.autograph.services.mozaws.net", "signingscript_geckoview_reference_browser_rel", "<%= scope.function_secret(["autograph_reference_browser_prod_password"]) %>", ["autograph_apk_reference_browser"], "autograph"]
     ],


modules/signing_scriptworker/templates/passwords-mobile.json.erb
```diff
@@ -11,6 +11,9 @@
     "project:mobile:fenix:releng:signing:cert:production-signing": [
         ["https://autograph-external.prod.autograph.services.mozaws.net", "signingscript_fenix_release", "<%= scope.function_secret(["autograph_fenix_production_password"]) %>", ["autograph_apk"], "autograph"]
     ],
+    "project:mobile:fenix:releng:signing:cert:fennec-production-signing": [
+        ["https://autograph-external.prod.autograph.services.mozaws.net", "signingscript_fennec_rel", "<%= scope.function_secret(["autograph_fennec_release_password"]) %>", ["autograph_apk_fennec_sha1"], "autograph"]
```
mitchhentges: Should the format be `autograph_apk_fennec_sha1` or `autograph_apk`? I chose `autograph_apk_fennec_sha1` so it would be the same as the "real" fennec signing config

JohanLorenzo: This is a great question! Thank you for asking it, otherwise I wouldn't have thought this through 🙂

When we switched to Autograph, we had to implement the SHA-1 support[1] because "old" Android couldn't be updated. What is "old" in this context? [2] reads:

>  * SHA256withRSA for minimum SDK 18 (Android 4.3) and up
>  * SHA256withDSA for minimum SDK 21 (Android 5.0) and up
>  * SHA256withEC for minimum SDK 18 and up
>  * SHA1with* for lower minimum SDK levels

The beta and release key is an RSA one, but the nightly key is a DSA one. Therefore, we may use SHA256 only if we target SDK 21 (so Android 5.0 [3]). That's confirmed by what people saw when we broke Nightly back then[4].

Good news: Fenix requires this SDK version[5] or above.

Having the `_sha1` suffix is just used to change the request[6] to ask Autograph to change the digest algorithm. It really changes it and not just adds it in addition to SHA-256[7].

Thus, I'm tempted to test the `autograph_apk` format and see what the signing_scriptworkers output.

What do you think @mitchhentges ?  

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1492245#c13
[2] https://jokester.io/post/2017-02/sha1-collision-and-apk-signing/
[3] https://developer.android.com/guide/topics/manifest/uses-sdk-element.html#ApiLevels
[4] https://bugzilla.mozilla.org/show_bug.cgi?id=1332916
[5] https://github.com/mozilla-mobile/fenix/blob/0fc42a8a0907d1581693bab8010779b7d4c1ab68/buildSrc/src/main/java/Config.kt#L18
[6] https://github.com/mozilla-releng/signingscript/blob/a6cee62177782a6626bc596bbb731aac03a4fe49/signingscript/sign.py#L946-L949
[7] Tested on https://download-installer.cdn.mozilla.net/pub/mobile/releases/68.1/android-api-16/multi/fennec-68.1.multi.android-arm.apk

mitchhentges: Sounds good, I'm down to roll `autograph_apk` and change it if there's an issue

PR #593, opened 2019-09-05
Install google chrome

dividehex:
This is untested at the moment and there might be some dependencies that need to be installed or upgraded.  I'll test before merging of course.


dividehex:
Tested on ubuntu t-linux64-ms

PR #597, opened 2019-09-10
Bug 1540276: Fix up other signing script workers r=jlorenzo

catlee:


catlee:
> https://github.com/mozilla-releng/build-puppet/blob/a4f6b97c3c193fd24f6cb9ea01846966d67c8237/modules/signing_scriptworker/templates/script_config.json.erb#L21
> 
> into quotes too. What do you think?

The code expected a json null value in some cases, which is why this is unquoted.

JohanLorenzo:
Got it. Thanks for the context! 🙂

PR #605, opened 2019-09-23
back out 663861a97adb1e8e4465b9ffd9506e5996438ece for potential bustage

escapewindow:


escapewindow:
Taskcluster's going to fail on this one, because these aren't latest.

lundjordan:
@escapewindow will pyup just pick this back up next week and reland along with any new version updates?

Or do we need to manually reland?

escapewindow:
It won't reland the `pyup: ignore` pins, or the beetmoverscript bump.

I've been debating whether to stop merging these, since a) they introduce potential bustage, and b) we're moving towards docker. If we want to keep merging these we should probably reland.

PR #611, opened 2019-09-30
Re-add myself back to shortlist so I can validate chain of trust for signing of mpd001

Callek:


escapewindow:
You should be able to verify cot via `verify_cot` anywhere, without needing access to the signing servers/scriptworkers. Is there something else you can’t verify?

PR #614, opened 2019-10-15
Bug1588596 worker monitoring

davehouse:
add cpu/temp/other monitoring on linux and macos workers


davehouse:
This one is running on one of the yosemite staging (394) and on 3 linux64 staging machines. The matching ronin-puppet change for mojave was tested on the -beta pool, but tests time results were not reviewed by qa/sheriffs (and timing is slowed enough we had to back it out yesterday on mojave).

PR #627, opened 2019-11-12
install zstandard pip for python 2 and 3

aerickson:
Also installs zstd CLI utility.

known issues:
- puppet < 4 doesn't support the same package name with two different providers, so exec must be used.

Testing done:
- converges locally in docker container
- converged successfully on t-linux64-ms-314 and t-linux64-ms-315
`puppet agent --test --server=releng-puppet2.srv.releng.mdc1.mozilla.com --environment=aerickson`

### DIFF
--- a/modules/packages/manifests/python2_zstandard.pp
+++ b/modules/packages/manifests/python2_zstandard.pp
@@ -5,10 +5,6 @@
 class packages::python2_zstandard {
     case $::operatingsystem {
         Ubuntu: {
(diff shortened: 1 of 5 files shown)


davehouse:
This looks great, and thank you for including the cli install. (TIL: github has a draft mode for pr's! I must have slept through that appearing)

aerickson:
PIPs have been uploaded. This is converging now.

PR #628, opened 2019-11-19
bug 1481916 - remove l10n bumper support.

escapewindow:
Non-urgent cleanup; I'm currently pinning both buildbot masters to my puppet env.


modules/l10n_bumper/manifests/services.pp
```diff
@@ -1,25 +0,0 @@
...
-
-    $env_config  = $config::l10n_bumper_env_config[$l10n_bumper_env]
-
-    $owner       = $users::builder::username
-    $basedir     = $l10n_bumper::settings::root
-    $python      = $packages::mozilla::python27::python
-    $mailto      = $l10n_bumper::settings::mailto
-    $config_file = $env_config['config_file']
-    file {
-        '/etc/cron.d/l10n_bumper':
```
Callek: Note removing this from the puppet manifest will still keep the file around on the host. So might be worth an `ensure=>` directive to remove it (or if you're just going to terminate/disable the instance we probably don't need the extra configuration)

modules/toplevel/manifests/mixin/l10n_bumper.pp
```diff
@@ -1,15 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-class toplevel::mixin::l10n_bumper {
-    include ::l10n_bumper
-
-    ssh::user_privkey {
-        "ffxbld_rsa":
```
Callek: same commentary about this key

modules/l10n_bumper/manifests/services.pp
escapewindow: Smart. I was planning on manually disabling and then cleaning up the puppet configs, and terminate the EC2 instances.

modules/toplevel/manifests/mixin/l10n_bumper.pp
escapewindow: I'll clean this up.

PR #629, opened 2019-11-20
Remove deprecated nodes

catlee:


manifests/moco-nodes.pp
```diff
@@ -229,391 +188,6 @@
...
 
-# Signing workers
-node /^signingworker-.*\.srv\.releng\..*\.mozilla\.com$/ {
-    $aspects       = [ 'maximum-security' ]
-    $only_user_ssh = true
-    include toplevel::server::signingworker
-}
-
-# Signing scriptworkers
-node /^signing-linux-\d*\.srv\.releng\..*\.mozilla\.com$/ {
```
escapewindow: We still use this for xpi, but it's pinned, so maybe we'll just keep that puppet env forked off an older rev.

manifests/moco-nodes.pp
```diff
@@ -229,391 +188,6 @@
...
-    $aspects                  = [ 'maximum-security' ]
-    $signing_scriptworker_env = 'prod'
-    $timezone                 = 'UTC'
-    $only_user_ssh            = true
-    include toplevel::server::signingscriptworker
-}
-
-node /^depsigning-worker.*\.srv\.releng\..*\.mozilla\.com$/ {
```
escapewindow: Same here.

manifests/moco-nodes.pp
```diff
@@ -229,391 +188,6 @@
...
-node /^tb-tree-comm-\d+\.srv\.releng\..*\.mozilla\.com$/ {
-    $aspects          = [ 'maximum-security' ]
-    $treescriptworker_env = 'tb-comm-prod'
-    $timezone         = 'UTC'
-    $only_user_ssh    = true
-    include toplevel::server::treescriptworker
-}
-
-## Loaners
-
```
escapewindow: maybe check with the related relops folk?

manifests/moco-nodes.pp
catlee: yeah, or just deploy directly to the worker. it will be moving to GCP soon, right?

escapewindow: +1

dividehex:
👍 That all looks good to me

PR #632, opened 2020-01-02
Remove mhentges and rgabas from jumphosts

klibby:
Rok is out, Mitch has changed teams.


davehouse:
I'll apply it on the jumphosts once it is merged.

PR #634, opened 2020-01-09
upgrade chrome on linux64 to 79

davehouse:


davehouse:
sparky confirmed with a test pool that the update is good

PR #671, opened 2020-06-19
Clean up moco nodes

dividehex:
This removes a lot of hosts that don't exist anymore.


dividehex:
> I’m guessing the Yosemite boxes are either gone or on ronin-puppet?
 This is correct.  Joel has confirmed they are no longer in use and therefore I am going to move them to ronin/mojave.

PR #687, opened 2020-10-07
Bug1656965 linux testers imagemagick, ffmpeg

davehouse:
This is not the portable/static build of imagemagick, but a copy of backports that tarek confirmed tests against.


aerickson (approved):
lgtm

PR #716, opened 2021-02-26
Allow access to hosts without using jumphost

dividehex:
In preparation of decommissioning the releng jumphosts, this PR allows ssh/vnc access directly from the corp vpn/jumphosts.  It covers all the hosts still managed under puppetagain.

### DIFF
--- a/modules/fw/manifests/profiles/bsdpy.pp
+++ b/modules/fw/manifests/profiles/bsdpy.pp
@@ -6,12 +6,12 @@ class fw::profiles::bsdpy {
 
     case $::fqdn {
         /.*\.mdc1\.mozilla\.com/: {
-            include ::fw::roles::ssh_from_rejh_logging
+            include ::fw::roles::ssh_from_anywhere_logging
             include ::fw::roles::nrpe_from_nagios
             include ::fw::roles::bsdpy_from_mdc1_releng
         }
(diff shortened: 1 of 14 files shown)


aerickson (commented):
syntax looks good. ping me if you'd like me to approve, deferring to releng and dhouse.

davehouse (commented):
I don't think I can be part to decide on this? (without something documented removing the security requirement for when we added the jumphosts)

escapewindow:
I'm mostly ok with this since I think we only have the testers in puppetagain.
However, I have questions about whether the decommissioning decision affects the vcs-sync and mac signer boxes. Maybe Releng needs more context about the decision and/or a quick sync up with Relops?

bhearsum:
If this means that additional people can route to the mac singers, I'm worried about scenarios where someone uses a 0-day to gain access without credentials. Eg: by hopping sideways from a VPN'ed machine they already have access to, or directly doing it through the VPN (assuming this change would allow that).

escapewindow (approved):
Covered in the IT meeting.
I assume we'll need to adjust our ssh configs?

bhearsum:
Was my concern discussed?

escapewindow:
> Was my concern discussed?

Discussed in Matrix. Looks like we're clear.

